PR #4, opened 2020-06-19
add react router and page components, fixes #2

### LINKED ISSUE #2: Enable routing
Need to enable moving between different urls in the app.

Use React Router (react-router-dom) to add routers.

HaSistrunk:
Fixes #2

I'd like a second pair of eyes on how we want to structure this and to make sure I'm setting this up right. Should these page components be in their own directory or subdirectory outside of or within `components`?


helrond:
@HaSistrunk I don't have strong opinions on structure at this point. Until we have a better sense of what components we need (and where those components get used), I'm not sure we can predetermine a structure. So maybe let's keep it simple for now and reorganize as we need to?

PR #24, opened 2020-07-08
Adds button component and smoke test

### LINKED ISSUE #7: Create Button component
A generic button component. 

Props to consider: 
- class
- aria-label
- icon
- type (button, submit, reset)
- href

helrond:
Interested in feedback on tests in particular - I feel like I'm fumbling around a lot in that area.

fixes #7


HaSistrunk:
For the icons, we'll be using Material Icons. I'm wondering if it's better to use inline SVG v. the icon font. I'm just thinking through how we'll want to hide the icons from screen readers, and what an icon-only button will look like. Happy to merge this as is and circle back as needed.

HaSistrunk (approved):

helrond:
I think that makes sense as an approach - likely this is something we'll need to look at in multiple components.

PR #29, opened 2020-07-13
Adds tile component and tests

### LINKED ISSUE #19: Create Tile component
Create component/s to handle sets of tiles, probably best represented as lists, in the DIMES search results and agent pages.

helrond:
fixes #19

### DIFF
--- a/src/components/Tile.js
+++ b/src/components/Tile.js
@@ -1,5 +1,4 @@
 import React, { Component } from "react";
-import PropTypes from 'prop-types';
 
 
 class HitCount extends Component {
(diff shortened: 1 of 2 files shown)


src/components/Tile.js
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from "react";
+import PropTypes from 'prop-types';
```
HaSistrunk:
```suggestion

```

HaSistrunk (approved):
I'm having trouble wrapping my head around this one. But I like the list approach, and we can always adjust things if needed

helrond:
I was following the docs here: https://reactjs.org/docs/lists-and-keys.html

PR #34, opened 2020-07-20
Add styles and restructures in component-specific folders

### LINKED ISSUE #31: Add scss files for components
The current components have no styling.

Add styling via SCSS for components. See a couple of informative posts:
https://www.sitepoint.com/react-components-styling-options/
https://scotch.io/tutorials/using-sass-in-create-react-app-v2
https://stackoverflow.com/questions/57829879/parcel-using-separate-scss-files-for-each-react-component-but-with-a-variables

## Additional context
Ultimately we'll want to make RAC SCSS importable as package, but for right now we can crib styles from the main RAC website.

helrond:
Grabs all SCSS from rockarch.org repository, does some minor restructuring to break into smaller components.

A couple of notes:
- typography has been moved to the typography stylesheet where mixins are typically created
- I've used BEM for CSS class names, or at least tried to. If you see any inconsistencies I'd be happy to know about them!

Moves each component into its own directory, each of which is structured as follows:
```javascript
Component/
  __tests__/
    Component.test.js   // component tests
  index.js    // the component 
  styles.scss   // SASS styles, usually just an import
```

fixes #31


src/styles/components/_buttons.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.btn--md  {
+  @include button-text;
+  border-radius: 3px;
+  font-size: 12px;
+  padding: 13px 14px 15px 14px;
+}
+
+.btn--lg {
+  font-weight: 700;
```
HaSistrunk, line 37: not sure which button this is, but just double checking that it doesn't contain text or a border radius.

HaSistrunk:
This looks great to me!

PR #44, opened 2020-07-29
Adds first draft of agent page

helrond:
Adds a first go at the PageAgent component.
- Adds a AgentAttribute and AgentAttributeList component to handle the pieces of data being displayed.

A couple of things are not done yet, and I'll open separate issues for those:
- Fetching data from Wiki* land. We don't have those identifiers in AS yet.
- Displaying related_agents. Likewise, this data is not in AS yet.


HaSistrunk:
This looks good! So exciting to get data in here. I bumped up the font size a little in the tiles to match the design treatment, and adjusted that tile type label color because the color contrast wasn't sufficient, if that's cool.

Is "Back to Search" actually a link that we're styling as a button? If I'm thinking about that right, I'm wondering if we should have a link component for that, or somehow refactor the skip link component to encompass that. Thoughts?

HaSistrunk:
And should this be limiting to 6 related collections yet? I'm not seeing that in my build

helrond:
OK, I added back some code which got reverted by mistake. If you pull down the most recent version the Argo development branch, you should now have collections limited to 6.

Regarding the "Back to Search" button - yes, that's a link. But so is the "Search More Related Collections", right?

HaSistrunk (approved):

Issue #49, opened 2020-08-07
Implement homepage search functionality


HaSistrunk:
Implement parameters from dropdown menu to limit search

PR #50, opened 2020-08-07
Create Homepage

### LINKED ISSUE #35: Build homepage
Build out PageHome.js

HaSistrunk:
Fixes #35

Adds basic components of the homepage without implementing search functionality, but still needs some work. I'm having a little trouble wrapping my head around using neat and taking the right approach to layouts. As it stands now, the hero is not responsive in the way it should be (elements are overlapping), and the search box has a non-responsive, fixed width. Can I get some help cleaning that up?

Since you are creating a menu component for the My List page, I held off adding that so as to not duplicate work.

It may make sense to make the hero its own component to be able to pull the styling out from layouts.scss.


helrond:
OK, there's still a bit more stuff to do here, but it's closer.
- Added form action and method so form now redirects as it should, fixes #49 
- Did a bit of futzing with the HTML structure and styling. On large screen sizes you'll see that the content is off-center, but that will be fixed when the changes I'm working on are merged (adding a `wrapper` div)

I think the dropdown in the search bar is really just a `select` input, right? The dropdown I created for the My List page is a different animal.

I think there's definitely opportunity to extract some more components, including the hero and possibly the search bar (since we'll want to use a similar component on the search page).

HaSistrunk:
Great, thanks for that cleanup and direction. I can keep working on this! And yes, you're totally right - that is a select input, not a menu.

HaSistrunk:
@helrond , I've cleaned this up a bit more:
- I pulled out search and hero components (the search component may need some style tweaking when we implement on other pages)
- Added the select input
- Cleaned up styles

What still needs to be done here or outside of this PR:
- I was struggling to get the button to display its full width. It's cutoff by the select and I'm not sure how to fix that.
- There's not yet any form validation
- Styling for the options in the select menu are apparently super limited. I wasn't able to make anything work yet, but was thinking we could handle in a separate issue
- Focus styles need some finessing, but I think this is something we can do across the site

helrond (approved):
This looks great! I've added a few commits which address the search button width and icon, but otherwise I think let's add issues for the other things and move on!

PR #55, opened 2020-08-10
Use span styled as table instead of line break

### LINKED ISSUE #33: Use styling to create non-semantic newlines in footer
Screen reader pauses at a new line, so when reading the heading "Rockefeller Archive Center" in the footer, it would pause between the first and second words because there is a `<br>` element there. The current `aria-label` is one way around that, but not ideal since this solution uses html for styling.

Replace `<br>` elements with styling to achieve a new line in the `<h2>` of the footer. 

If we decide to make this change, we can remove the aria-label. If we decide not to make the change, we can leave as-is.

## Additional context
[https://css-tricks.com/injecting-line-break/](https://css-tricks.com/injecting-line-break/)

helrond:
fixes #33

### DIFF
--- a/src/components/Footer/index.js
+++ b/src/components/Footer/index.js
@@ -8,7 +8,7 @@ class Footer extends Component {
       <footer className="footer" role="contentinfo">
         <div className="footer-primary">
           <div className="container">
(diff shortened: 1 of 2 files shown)


HaSistrunk:
Yeah, this seems to work! I tested with Voiceover on a few browsers.

HaSistrunk (approved):

PR #53, opened 2020-08-10
Add MyList page component and associated components

helrond:
Adds the `PageMyList` component. Also adds a `Dropdown` and `SavedItemList` Components. This PR does _not_ include modals or their onClick handlers; that will be left for a yet-to-be-filed issue and PR.

@HaSistrunk can you take a look at the implementation in the `Dropdown` component - I think I might be missing some accessibility attributes.

fixes #38 
fixes #11 
fixes #12 
fixes #20

PR #56, opened 2020-08-10
Implement icon component

### LINKED ISSUE #51: Implement Material icons
Determine if we should use font or svg icons. Implement in a way that hides them from screenreaders, and so that they do not  receive keyboard focus (if svg).

helrond:
Implements Material Icons as a separate component. This should be merged AFTER #53 has been merged to `base`.

fixes #51

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: node_js
+node_js:
+  - "stable"
+cache:
+  directories:
+    - node_modules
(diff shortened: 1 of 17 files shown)


HaSistrunk:
Looks great! I switched the `<i>` to `<span>` elements for the icons just for semantics, if that's cool with you, although it doesn't functionally matter.

PR #73, opened 2020-08-21
Allow users to send emails with data about records

### LINKED ISSUE #13: Create Modal component
An accessible modal component that manages keyboard focus.

## Additional context
Consider using existing module [react-modal](https://github.com/reactjs/react-modal) or see other examples:
1. [Reach-UI modal](https://reacttraining.com/reach-ui/dialog)
2. [Instructure Design modal](https://instructure.design/#Modal)

### LINKED ISSUE #71: Allow users to email list
Users need to email lists to themselves and/or others.

Implement the Components and functions necessary for this feature request:
- A modal with the necessary fields
- onClick handlers to load modal with data
- functions to interact with request broker endpoints

## Design reference
Screen Shot 2020-08-19 at 6 31 59 PM

helrond:
Adds Modal component, fixes #13 

Adds EmailModal, along with handlers for request broker, fixes #71

NOTE: the ReCAPTCHA site key used here is for testing purposes only, and results in some UI ugliness. This will be resolved once we are closer to production.

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -1,0 +1,147 @@
+import React, { Component } from 'react';
+import ReCAPTCHA from "react-google-recaptcha";
+import PropTypes from "prop-types";
+import Modal from 'react-modal';
+import Button from "../Button";
+import {TextInput, TextAreaInput} from "../Inputs";
+import MaterialIcon from "../MaterialIcon";
+import {ModalSavedItemList} from "../SavedItem";
+import "./styles.scss"
+
+Modal.setAppElement("#root");
+
+class MyListModal extends Component {
+  render() {
+    return (
+      <Modal
+        isOpen={this.props.isOpen}
+        className="modal-content"
+        overlayClassName="modal-overlay">
+        <div className="modal-header">
+          <h2 className="modal-header__title">{this.props.title}</h2>
+          <button className="modal-header__button" onClick={this.props.toggleModal}>
+            <MaterialIcon icon="close"/>
+          </button>
+        </div>
+        <div className="modal-body">
+          <div className="modal-list">
+            <ModalSavedItemList items={this.props.list} />
+          </div>
+          <div className="modal-form">
+            <form>
+              <div className="modal-form__input-group">
+                {this.props.inputs}
+              </div>
+              <div className="modal-form__captcha">
+                <ReCAPTCHA
+                  sitekey="Your client site key"
+                  onChange={this.handleCaptchaChange} />
+              </div>
+              <div className="modal-form__buttons">
+                {this.props.buttons}
+              </div>
+            </form>
+          </div>
+        </div>
+      </Modal>
+    )
+  }
+}
+
+MyListModal.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  toggleModal: PropTypes.func.isRequired,
+  handleCaptchaChange: PropTypes.func.isRequired,
+  title: PropTypes.string.isRequired,
+  list: PropTypes.array.isRequired,
+  inputs: PropTypes.node.isRequired,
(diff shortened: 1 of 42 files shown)


src/components/SavedItem/index.js
```diff
@@ -51,11 +51,34 @@ SavedItem.propTypes = {
...
+      <div className="modal-saved-item">
+        <div className="modal-saved-item__inputs">
+          <CheckBoxInput
+            className="checkbox--orange hide-label"
+            id={this.props.title}
+            checked={true}
+            label={this.props.title} />
+        </div>
+        <div className="modal-saved-item__item-description">
+          <h3 className="modal-saved-item__title">{this.props.title}</h3>
```
HaSistrunk: I'm looking at headings in the modal, and wondering if these should just be a list of checkboxes? On the My list page they have more content under them, but on the modal it's just the titles.

helrond: Hmm, good call - all those divs are probably unnecessary...

src/components/SavedItem/index.js
```diff
@@ -88,8 +109,35 @@ SavedItemGroup.propTypes = {
...
+      <ModalSavedItem
+        key={index}
+        title={item.title}
+        uri={item.uri} />
+    );
+  }
+  render() {
+    return (
+      <div className="modal-saved-items__item-group">
+        <h2 className="modal-item-group__title">{this.props.title}</h2>
```
HaSistrunk:
```suggestion
        <h3 className="modal-item-group__title">{this.props.title}</h3>
```

HaSistrunk:
Yeah, this looks good to me! I just added functionality to esc or click out of the modal to close it, and a button label on the close button.

Did you want to talk about any of the TO DO questions, or deal with those later?

One thing I'm seeing in the My List page is that the `Remove` button is floating away. I may have missed that in the last PR, because I don't see that you did anything to introduce that here.

HaSistrunk:
One other little thing: looking at the modal form `<div>`, I'm seeing the that light-grayish-yellow background color of the `modal-form` class doesn't extend the full height of the modal.

helrond:
OK, all the issues you raised should be addressed. I also did a few things globally:
- Converted components that don't use state to function-based components, which helps to remove some lines of code.
- Used better destructuring methods to. make functions more readable.s
- Add some missing tests for components.

The one TODO I'd be curious about hearing your thoughts on is whether the buttons in the modal form should more properly be inputs.

src/components/Modal/index.js
```diff
@@ -0,0 +1,171 @@
...
+  toggleModal: PropTypes.func.isRequired,
+  handleCaptchaChange: PropTypes.func.isRequired,
+  title: PropTypes.string.isRequired,
+  list: PropTypes.array.isRequired,
+  inputs: PropTypes.node.isRequired,
+  buttons: PropTypes.node.isRequired
+}
+
+export class EmailModal extends Component {
+  // TODO: should buttons be inputs instead?
```
HaSistrunk, line 62: I don't think so. [Aren't they functionally the same](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event)? My research tells me it can be harder to style an input to look like a button across all browsers.

HaSistrunk:
Nice. Ok, this all looks good to me. I just want to note that our current pattern is to have the button disabled until the reCAPTCHA is checked, but not to tie the button being disabled to the required email address field being filled out. I like the idea of using form validation and error messages instead of disabling buttons so that users can try to submit something and get a message telling them what the problem is, rather than preventing them from submitting and trying to figure out why they can't. Also, disabled buttons don't receive keyboard focus. I think this is something we can worry about when setting up form validation.

HaSistrunk (approved):

PR #74, opened 2020-08-21
Improve responsiveness of top nav

### LINKED ISSUE #32: Improve mobile top nav behavior
Top nav should collapse for phones.

Add necessary HTML, CSS and JS to make this nav behave as mocked up.

helrond:
Moves all dropdowns into Dropdown component and adds a NavDropdown for mobile nav.

Note: the design treatments calls for behavior which pushes the main content left when the menu is open. I did not implement this because it seemed like that would require adding `margin-left` and `margin-right` properties on `#root`, and my brain couldn't wrap itself around how to do that.  I'm going to file an issue, but I suspect this will likely require setting state in `App.js` and then passing that down to the menu via props.

fixes #32

### DIFF
--- a/src/components/Nav/index.js
+++ b/src/components/Nav/index.js
@@ -1,13 +1,16 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import MaterialIcon from "../MaterialIcon";
+import {NavDropdown} from "../Dropdown";
 import "./styles.scss";
 
 export class NavItem extends Component {
   render() {
     return (
-      <li className={`nav__item ${this.props.className}`}>
-        <a className={`nav__link ${this.props.className}`} href={this.props.href}>{this.props.label} {this.props.icon && <MaterialIcon icon={this.props.icon} />}</a>
+      <li className={`nav__item ${this.props.className ? this.props.className : ""}`}>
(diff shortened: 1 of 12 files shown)


src/components/Nav/index.js
```diff
@@ -16,10 +18,11 @@ NavItem.propTypes = {
 }
 
 export const Nav = ({ ariaLabel, children, className}) => (
-  <nav className={`nav ${className && className}`} aria-label={ariaLabel}>
-    <ul className="nav__list">
+  <nav className={`nav ${className && className} : ""`} aria-label={ariaLabel}>
```
HaSistrunk: @helrond can you take a look at this again? I'm not sure I pulled this together correctly when fixing the merge conflicts with base with the switch from class to functional component. I'm also seeing that the My List link comes and goes at different screen sizes before the Dropdown kicks in.

helrond: Yeah that should be 
```
{`nav ${className ? className : ""}`}
```

In terms of the nav, I think there may be an issue with the design where the lower end of the break points isn't accounted for. I may have to tweak padding/margins or something.

HaSistrunk (approved):

PR #79, opened 2020-08-25
Adds reading room request modal and handlers

### LINKED ISSUE #70: Allow users to request records in reading room
Users need to be able to request desired records for viewing in the reading room.

Implement the Components and functions necessary to support this feature:
- A modal with necessary fields
- onClick handlers for buttons
- functions to interact with Request Broker as needed

## Design reference
Screen Shot 2020-08-19 at 6 29 07 PM

helrond:
Like the other modals, this doesn't fully work without the request broker yet, but there should be enough here to see what's going on.

Fixes #70

### DIFF
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -41,7 +41,7 @@ const Dropdown = (props) => {
   )
 }
 
-export const MyListDropdown = ({ removeAllItems }) => (
+export const MyListDropdown = ({ readingRoomRequest, removeAllItems }) => (
   <Dropdown
     label="Actions"
     iconBefore="settings"
@@ -59,7 +59,7 @@ export const MyListDropdown = ({ removeAllItems }) => (
         {
           "label": "Request in Reading Room",
           "iconBefore": "local_library",
-          "handleClick": null
+          "handleClick": readingRoomRequest
         },
         {
           "label": "Request Copies",
--- a/src/components/Inputs/index.js
+++ b/src/components/Inputs/index.js
@@ -47,13 +47,14 @@ export const DatePickerInput = (props) => (
     <input
       type="date"
       id={props.id}
+      aria-describedby={`desc-${props.id}`}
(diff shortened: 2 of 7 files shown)


HaSistrunk:
I'm looking at this now, and one thing I wanted to flag was that element ids can't have whitespace, so our input ids and associated labels aren't functioning properly where they're pulling in the item titles.

helrond:
@HaSistrunk can you point me at an instance of this?

HaSistrunk:
> @HaSistrunk can you point me at an instance of this?

For example for the ModalSavedItem, it's using `{title}` for the `id` and `label`:
```
const ModalSavedItem = ({ title }) => (
  <li className="modal-saved-item">
      <CheckBoxInput
        className="checkbox--orange"
        id={title}
        checked={true}
        label={title} />
  </li>
)
```

Since `{title}` is pulling from the data fetched in the PageMyList component, these have whitespace. So when I inspect the page I see something like this:

```
<div class="checkbox--orange">
    <input type="checkbox" id="Meade Alcorn Interview" name="Meade Alcorn Interview" checked="">
    <label for="Meade Alcorn Interview">Meade Alcorn Interview</label>
</div>
```

helrond:
@HaSistrunk got it! Let me fix that.

HaSistrunk:
Ok, just a few things:

1. I added an aria-describedby to connect the input and the help text in the forms. 
2. This is my fault from a previous PR I think, but our headings are a little funky because our item group title needs to be an h2 on the My List page, and an h3 in the modal (since the modal title is an h2). I made a change for that.
3. I can't get the date picker to open with the keyboard

helrond:
@HaSistrunk Thanks for these edits. I've added an issue to track the last thing you raise (accessibility of date inputs), as there doesn't appear to be an easy fix for this (or even consensus on what the problem is).

HaSistrunk (approved):

PR #82, opened 2020-08-26
Improve button styling

### LINKED ISSUE #58: Improve button styling
Button icons and labels aren't aligned, and there is a size discrepancy between buttons, likely related to some having borders and some not.

## Additional context
image

HaSistrunk:
- Improves button icon alignment and sizes
- Adds hover style to mobile nav button

Question: `.btn--mobile` and `.btn--mobile-dropdown` are not currently being used. Should we delete, or hold on for a future use?

Fixes #58

### DIFF
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -89,7 +89,7 @@ export const NavDropdown = () => (
   <Dropdown
     iconBefore="menu"
     iconBeforeOpen="close"
-    className="nav-mobile hide-on-lg-up"
+    className="hide-on-lg-up"
     buttonClassName="btn nav-mobile__btn"
     itemClassName="btn--navy btn--mobile-dropdown dropdown__item--navy"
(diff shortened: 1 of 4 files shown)


helrond:
Cool, this looks good. A couple of things:
- I'm seeing `.btn--mobile-dropdown` being used, but not `.btn--mobile`. I don't see styles for the latter class though...
- The hover for the mobile button is a little wonky. I think the margin on the wrapping div might need to be swapped for padding, and a better height applied. I'm also not convinced we need a hover state, since only mobile devices will see that menu.

HaSistrunk:
Thanks @helrond, you're right abut the `.btn--mobile` classes, I'll leave them alone.

I made some slight changes to enable and standardize hover on the nav buttons - think it's nice to have in case someone shrinks the window on a desktop, maybe?

helrond (approved):
Cool, I just changed the border color on hover and removed the border-radius from the mobile nav button. All good to go!

PR #84, opened 2020-08-31
Adds form validation

### LINKED ISSUE #18: Research form implementation
Research the best approach to implementing forms in React, and decide if a solution like [Formik](https://jaredpalmer.com/formik/docs/overview) is appropriate.

### LINKED ISSUE #65: Implement form validation
Add form validation and error messages to components with forms, including those in the My List modal, and the Search component.

### LINKED ISSUE #77: Handle change when checkboxes in MyList (including modals) are changed
Checking or unchecking a one of the checkboxes in MyList, including the modals, should result in that item being added to or removed from the list that is submitted.

Add function(s) which handle this action.

### LINKED ISSUE #87: Connect My List selected items to modal items
When a user unchecks checkbox inputs on the My List page for items, and then chooses an action such as "Request in Reading Room", "Request Copies", or "Email List" for those items, only the checked items should show up in the modal. Otherwise, they have to unselect unwanted items multiple times.

Either there should be no option to uncheck items on the My List page (ie. no checkbox inputs), or the unchecked items should not appear in the modal items list.

Sorry if I'm missing specifics about the mechanisms by which this is functioning. Hopefully this is clear.

helrond:
Implements [Formik](https://formik.org/) for form validation, which also helps us clean up a bunch of other things.

On the accessibility side of things, I've implemented functionality to focus on an element which has an error message if one exists when forms are submitted. I've also added aria attributes for errors and required fields. I did not implement focus on the first field in the form because I'm not sure if that's actually what users want in this case.

I've also added handlers for checkbox actions - the way this works now is that items which are unchecked in the main page are removed from the list which shows up in the modals. Then, in the modals, items which are unchecked there are removed from the list to be submitted (tied to a hidden input in the form).

Fixes #65 
Fixes #18 
Fixes #87 
Fixes #77


src/components/Form/index.js
```diff
@@ -0,0 +1,80 @@
...
+        tabIndex="0"
+        className={ errors && errors[name] && touched[name] ? "is-invalid" : null }
+        type={type}
+        name={name}
+        id={name}
+        component={component}
+        rows={rows}
+        children={children}
+        aria-invalid={errors && errors[name] && touched[name] ? 'true' : null}
+        aria-describedby={errors && errors[name] && touched[name] ? `${name}-error` : null}
```
HaSistrunk: Can we add `desc-${name}` as an initial aria-describedby value that corresponds to the `helpText`, and then when there is an error, add this additional `${name}-error` value? I couldn't figure out how to get both.

helrond: I think that would look something like:
```
{errors && errors[name] && touched[name] ? `${name}-error desc-${name}` : desc-${name}}
```
The potential issue I see here is that there isn't always help text, so this may in some cases point to a non-existent DOM element.

HaSistrunk (approved):

PR #89, opened 2020-09-02
Adds SVG social media icons

### LINKED ISSUE #88: Add social icons
Social icons are not displaying, because they rely on Font Awesome.

Use SVG files for icons.

helrond:
Adds SVG icons for social media feeds. Fixes #88


HaSistrunk:
Apparently these can be squirrelly in different ways with various SR + browser combos. There are various approaches, but my research tells me that adding `aria-hidden="true"` and `focusable="false"` to the `<svg>`, and then just adding an `aria-label` with text for the sr on the anchor tag (instead of needing the visually hidden span) is a well-supported approach.

HaSistrunk (approved):

helrond:
Thanks @HaSistrunk!

PR #99, opened 2020-09-11
Adds facet modal

### LINKED ISSUE #91: Add modal with facet filters
Users want to filter results by certain categories.

Add a modal which presents filter options, and filter results based on actions in that modal.

helrond:
Adds a facet modal which filters search results.

Some implementation notes:
- For most of the facets, changes in values trigger an updated search. The exception to this is the dates filter, which is filtered when a user clicks "apply"
- Making a change in a facet does not (right now) close the facet modal.

fixes #91

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,2 @@
-REACT_APP_ARGO_BASEURL=http://10.0.1.90:8010
+REACT_APP_ARGO_BASEURL=http://localhost:8000
 REACT_APP_REQUEST_BROKER_BASEURL=http://request-broker.rockarch.org
--- a/src/components/Facet/index.js
+++ b/src/components/Facet/index.js
@@ -3,22 +3,31 @@ import { CheckBoxInput } from "../Inputs";
 import "./styles.scss";
 
 
-const FacetItem = ({ count, handleChange, label }) => (
+const FacetItem = ({ checked, count, handleChange, label, paramKey }) => (
   <CheckBoxInput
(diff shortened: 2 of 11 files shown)


HaSistrunk:
Cool, this looks great. I'm getting some weird behavior with the creator checkboxes. When I select one, the facets disappear in the modal/there are no search results.

HaSistrunk:
> Cool, this looks great. I'm getting some weird behavior with the creator checkboxes. When I select one, the facets disappear in the modal/there are no search results.

I think that is just a data pipeline thing, no issue with the code that I can see.

HaSistrunk (approved):

helrond:
@HaSistrunk can you describe the exact steps which produced this behavior so I can see if I can reproduce?

HaSistrunk:
Initially, when I did a search for "Rockefeller" and tried to facet by Geither, which appeared in the facets as having 1 record, there were no results. However, after I pulled in the most recent version of Argo, that is no longer happening and it is giving the expected results.

However, if I select any of the creator or subject inputs, it gives me only one search results even if the number next to the name indicates more than one results, such as "Bergmann, M. (Max) (3)"

helrond:
OK cool. Yeah it will be interesting to see what users make of this - grouping hits by collection means that you get some differences in the number of "hits" versus the number of tiles in the list. In the case you mention, there are three hits within a single collection.

Issue #97, opened 2020-09-09
Improve accessibility of search results tiles


HaSistrunk:
## Is your feature request related to a problem? Please describe.
Users should be able to click on the tile to link to the collection details, but including the entire contents of the tile as part of the link makes for a very verbose link for screen readers, and may lead to confusion because the control to open the collection results modal is inside the link.

## Describe the solution you'd like
One option would be to use the pattern [described here](https://inclusive-components.design/cards/) as the pseudo content trick to make the `h2` the link, and use css to stretch the link's layout over the whole tile. While the "# matches" button currently comes before the title in the DOM, it should ideally receive focus after the `h2` title.

## Additional context
It may make sense to see how the current pattern holds up for mouse users in user testing before tackling this.

HaSistrunk:
- Pad out the link to encompass the whole tile
- Update the DOM order so that elements in the tile are all after the heading

PR #102, opened 2020-09-14
Adds collection results modal

### LINKED ISSUE #92: Add collection results modal
Users want to see groups of hits in a given collection.

Add a modal which shows the first level of children and groups of hits within them. This will likely require the development of an endpoint in Argo see RockefellerArchiveCenter/argo#71

helrond:
Adds modal which shows hits within a collection. I made some slight adjustments to the tile HTML, which may start to address #97 

Also uses custom CSS classes for styling modals.

NOTE: Branched from `issue-100` so that should be merged first.

fixes #92

### DIFF
--- a/src/components/CollectionHits/index.js
+++ b/src/components/CollectionHits/index.js
@@ -1,0 +1,30 @@
+import React from "react";
+import { HitCount } from "../Tile";
+
+const CollectionHitsInfo = ({collection}) => (
+  <div className="collection-hits__info">
+    <h4 className="collection-hits__title">Collection Name:</h4>
+    <a className="collection-hits__link" href={collection.uri}>{collection.title}</a>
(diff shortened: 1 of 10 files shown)


HaSistrunk:
Cool! We already have this flagged for usability testing, but just noting two things to track as we test and solidify an approach here:
1. We should probably remove the tile hover style if the collection title is the thing the user needs to click to access to link.
2. It does not seem obvious to me what is and is not a link in the hits modal until I hover/focus on them. We may need more consistent or obvious styles there, but we'll see.

HaSistrunk (approved):

helrond:
Yup, those are *definitely* things we want to test. I remain unconvinced that anyone will click on that matches button expecting something different to happen than just going to the collection.

PR #113, opened 2020-09-17
Improves accessibility of tiles

### LINKED ISSUE #97: Improve accessibility of search results tiles

helrond:
Puts elements in desired focus order, makes entire tile linked. Fixes #97 

I also addressed the order of elements in the collection hit display.

### DIFF
--- a/src/styles/components/_tile.scss
+++ b/src/styles/components/_tile.scss
@@ -14,7 +14,8 @@
 
 .tile {
   @include grid-column(12);
-  display: inline-block;
+  display: flex;
+  flex-wrap: wrap;
   box-sizing: border-box;
   background: $off-white;
   border: 1px solid $orange-gray;
@@ -22,6 +23,7 @@
   border-radius: 4px;
   padding: 26px 20px;
   margin-bottom: $gutters-default;
+  position: relative;
   &:hover {
     border: 2px solid $red-orange;
     padding: 25px 19px;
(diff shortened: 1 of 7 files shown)


src/styles/components/_tile.scss
```diff
@@ -101,19 +111,18 @@
 
 .tile__hit-count {
   @include tile-hit-count-text;
-  float: right;
+  position: relative;
+  order: 1;
+  top: -32px;
+  right: -70px;
```
HaSistrunk:
```suggestion
  position: absolute;
  order: 1;
  top: 0px;
  right: 20px;
```

HaSistrunk: The tile hit count positioning is a little off. I think this fixes it?

HaSistrunk: The tile hit count button positioning is off. This doesn't entirely fix it - on hover it's still wonky

helrond: Yeah this is a function of the border width change on hover state. My googling has not yet revealed a solution for this, although I am sure there is one somewhere!

HaSistrunk: Happy to wait and deal with this later if we need to .

HaSistrunk (approved):

Issue #122, opened 2020-09-29
Provide expanded ancestor view at URL referencing an object


helrond:
## Is your feature request related to a problem? Please describe.
When users load a URL, they should see the requested object/collection in context of its ancestors.

## Describe the solution you'd like
Display the expanded tree of ancestors with the object in context. 

This is related to #121 and #120

Issue #121, opened 2020-09-29
Create complete ancestor display


helrond:
## Is your feature request related to a problem? Please describe.
Archival objects do not display ancestors above their parent resource record, including resources included in an arrangement map.

## Describe the solution you'd like
Provide a mechanism for displaying all ancestors. Options include:
1. Add a complete ancestor tree at the time of indexing. The main downside to this is substantially increased pre-processing time.
2. Add only the parent ancestor at the time of indexing. This would require the app to iterate up the ancestor tree, fetch data about each ancestor in turn.
3. Provide an endpoint which returns ancestors. This might be a way of splitting the difference on the pre-processing and rendering time.

Issue #120, opened 2020-09-23
Handle URL changes for clicks within Collections Context


helrond:
## Is your feature request related to a problem? Please describe.
Users should be able to use the browser back/forward buttons to navigate their browser history within collection context. 

## Describe the solution you'd like
- The URL should also update (ideally without page reload) as they click around in the collection context.
- Forward and back buttons should follow the user's navigation history
- Search parameters should be persisted throughout

PR #123, opened 2020-09-29
Implements collection/object context

### LINKED ISSUE #111: Build out collection context component
Build out collection context component as specified in design treatments.

### LINKED ISSUE #112: Build out object context component
Build out the object context component as specified in design treatments.

### LINKED ISSUE #118: Collection Hits modal can be slow to respond
## Describe the bug
Depending on the size of the collection, the collection hits modal can be slow to open.

## Expected behavior
Collection hits modal should open right away, with an indication that content is loading.

helrond:
Adds a collection/object context panel, which can be used to drill down into collections, fixes #111 and fixes #112.

Also shows collection hits modal immediately and uses loading skeletons to indicate loading of children, fixes #118.

There are several outstanding pieces of this functionality, which have been broken out into their own issues:
- #120 Handling URL changes as users click around in the tree
- #121 and #122 Better ancestor display

In the interest of maintaining forward velocity, it may be useful to get this merged in and deal with those issues separately.


HaSistrunk:
A few things (which can be made into separate issues depending on how much you want to focus work in this PR):
1. Does this also fix #118 ?
2. Toggle buttons on mobile:
    a. Let's choose either "Collection Content" or "Collection Context" and match between the heading and button text?
    b. What do you think about changing the tabindex values to change the focus order for these buttons? I think it would make sense to have them come before the content instead of after it.
4. We need some styles for keyboard navigation to show focus on the collection context list items
5. I'm thinking we should use aria live regions to announce the changes to the details when a user selects a collection or object. I created issue #125 for that.
6. I'm doing some research into the html structure/accessibility for the collection context with the nested lists and buttons. I'll follow up.
7. The "NOTE TO RESEARCHER" in this collection is not in a description in the details. Is that because of the type of data it is? 
image

HaSistrunk:
PS, this is really exciting and looks great!

helrond:
Thanks for the thorough check @HaSistrunk! I'm marking this as ready for review.

To respond to your questions directly:
1. Yes! I've added that into the PR description
2. Good call, I've gone with "content" across the board (which involved renaming some components and props as well) and moved the buttons in the DOM (since they have a fixed position it seemed easier to do that than futz with tab index)
3. Added `:focus-within` styles to list items.
4. Awesome, thanks for adding this issue
5. Cool - in the later stages of working on this it occurred to me that I was basically building an accordion, so that may be a good frame to think about this from.
6. Good catch - I am going to file an issue about this one because right now we need to get note types into Elasticsearch so we can more reliably grab them. Right now we are relying on titles, which is...prone to failure. In this case we were looking for "Scope and Contents Note" but the title is "Scope and Contents note".

HaSistrunk:
GREAT!
5. As discussed, an accordion where multiple things can be open at the same time would improve this pattern. I'll add a separate issue to tackle that change.

HaSistrunk (approved):

PR #124, opened 2020-09-30
Displays full ancestor trees in detail component

### LINKED ISSUE #121: Create complete ancestor display

helrond:
Takes advantage of new `/ancestors/` detail endpoint to display full ancestor tree in RecordsDetail component. This also allows for faster response times as ancestors are fetched and loaded separately from the collection description.

An example of a deeply nested object: objects/VgDT39ox36DWutz9eduq6k

Fixes #121


HaSistrunk (approved):
Yep!

PR #126, opened 2020-09-30
Adding box icon

### LINKED ISSUE #116: Add SVG icon for box
We are using Material Icons for the icons on DIMES, but this library does not appear to have a "box" icon. It look like that was pulled from FontAwesome.

It would be dumb to import FontAwesome for just one icon, so let's see if we can find an SVG file for this and use that instead, like we did for the social icons.

helrond:
Fixes #116


HaSistrunk (approved):
Nice. The subcollection icons look large now in comparison, but not a dealbreaker.
image

helrond:
@HaSistrunk I can fix that - there's a hardcoded size in the SVG that needs to be tweaked. Now that all the code is in place I can make that look pretty.

PR #138, opened 2020-10-05
Implements "Add to list" buttons in collection content

helrond:
Implements "add to list" buttons in collection content component. Moves the `savedList` state up to the top of the app, since three different components need access to that data.

fixes #132

PR #139, opened 2020-10-05
Ties modal submit buttons to submitList prop

### LINKED ISSUE #137: Item number does not change in Request Item(s) button
## Describe the bug
The `ReadingRoomRequestModal` and `DuplicationRequestModal` form button label text does not change to match the number of items in the submit list.

## To reproduce
In a request modal, deselect an item checkbox input. Notice that the form button text does not change.

## Expected behavior
The number on the label should change to match the number of checked items. The word "items" should also put the "s" in parenthesis to match the case of having only 1 item: "Request 1 item(s)"

helrond:
This means the count of items will always be accurate in submit buttons. Also added logic to pluralize the "item" based on the length of that same array. Fixes #137

NOTE: This is branched from #138, so merge that first!


HaSistrunk:
Oh, getting fancy with the "s"!

- So right now if I open a modal and uncheck a couple items, the button label changes to match the submitList. But then if I close that modal and open it/any modal with a request number again, the button label has the number that it did before regardless of how many items are checked in the current modal.
- What do we do with 0 items in the submitList (that may be a separate question from this PR)?

helrond:
OK cool - I took care of the first thing. The second (no items in the list) should probably go in a separate issue so we can discuss implementation there. My initial thought is to include that in form validation.

src/components/Modal/index.js
```diff
@@ -190,7 +191,7 @@ export const ReadingRoomRequestModal = props => (
             component="div"
             className="modal-form__error" />
           <FormButtons
-            submitText={`Request ${props.submitList.length ? (props.submitList.length) : ""} ${props.submitList.length > 1 ? "Items" : "Item"}`}
+            submitText={`Request ${props.submitList.length ? (props.submitList.length) : "0"} ${props.submitList.length > 1 ? "Items" : "Item"}`}
```
HaSistrunk:
```suggestion
            submitText={`Request ${props.submitList.length ? (props.submitList.length) : "0"} ${props.submitList.length !== 1 ? "Items" : "Item"}`}
```

src/components/Modal/index.js
```diff
@@ -299,7 +300,7 @@ export const DuplicationRequestModal = props => (
...
```
HaSistrunk:
```suggestion
            submitText={`Request ${props.submitList.length ? (props.submitList.length) : "0"} ${props.submitList.length !== 1 ? "Items" : "Item"}`}
```

src/components/Modal/index.js
HaSistrunk: So that it's 0 items, 1 item, and >1 items

HaSistrunk (approved):

PR #145, opened 2020-10-06
Adds custom checkbox styling

### LINKED ISSUE #143: Improve styling of checkboxes
Checkboxes require custom styling to properly set the background colors.

Add custom styles for orange and blue checkboxes.

helrond:
Adds custom checkbox styles with colors that match RAC theming.

@HaSistrunk Can you look at this from an accessibility perspective, particularly what I'm doing on the MyList page? This approach involves messing with the input label, which makes me feel a little queasy. Other approaches I've seen use a different input and style it to look like a checkbox, which also seems wrong. Bottom line - if this is an accessibility problem I don't have an issue with having non-pretty checkboxes in this application.

fixes #143


helrond:
Also I'm wondering if there's a better way to do the checkboxes on the main My List page. Right now there's a checkbox and then a div containing description that's visually adjacent but is otherwise not tied to that checkbox. I'd love to be able to use the label attribute in that description div, but we do need other data besides just the title in that div.

HaSistrunk:
Since we got rid of the checkboxes on the My List page, there's no issue there. If we end up adding those back in we can think through that a little more. This seems to be working ok as is, though. I made the checkmark a little smaller for aesthetics and fixed a problem keeping the digital material facet checkbox from appearing by adding a class.

HaSistrunk (approved):

helrond:
Great, thanks! It's nice when fixing one problem solves another...

PR #146, opened 2020-10-06
Opens fee schedule in new window

### LINKED ISSUE #144: Make links in modals open in a new window
Request copies modal has a link to "fee schedule" that takes user to the rockarch.org Access and Request Materials page. However, if a user has already started filling out the request form and then clicks this link, they will lose the info they have entered in the form when the browser loads the new page.

- Open the "fee schedule" link in a new tab.
- For accessibility purposes, add attribute `title="opens in a new window"` to the link.

helrond:
Uses `target="_blank"` to open link in new tab or window. Also adds `title` attribute for accessibility and `rel=noopener` as recommended [here](https://mathiasbynens.github.io/rel-noopener/).

fixes #144

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -283,7 +283,14 @@ export const DuplicationRequestModal = props => (
             component="textarea"
             rows={5} />
           <FormGroup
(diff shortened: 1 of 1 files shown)


HaSistrunk (approved):
Perfect! Good call to update the link to go right to the relevant heading.

PR #150, opened 2020-10-07
Implements accordion to improve accessibility

### LINKED ISSUE #133: Use Accordion to Present Collection Content
Explore replacing the nested lists with an accordion approach to improve accessibility.

helrond:
Replaces existing `ul` with `react-accessible-accordion` component.

Only collections (i.e. things that have children and are therefore expandable) are Accordion buttons. Objects are simple divs with a button that's padded out.

fixes #133

### DIFF
--- a/src/components/RecordsContent/index.js
+++ b/src/components/RecordsContent/index.js
@@ -49,7 +49,8 @@ class RecordsChild extends Component {
   }
 
   render() {
-    const {handleObjectClick, item, params, savedList, setActiveRecords, toggleInList, toggleIsLoading } = this.props;
+    const { ariaLevel, handleObjectClick, item, params, savedList,
+            setActiveRecords, toggleInList, toggleIsLoading } = this.props;
     return (item.type === "object" ?
       (<div className={`child__list-item child__list-item--${item.type} ${item.isActive ? "active" : ""}`} >
         <button className={`child__title child__title--${item.type}`} onClick={() => handleObjectClick(item)}>{item.title}</button>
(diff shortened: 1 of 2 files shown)


HaSistrunk:
@helrond can you do a pass to look at those lines on the left that link records as things are expanded? I started to work on this but wasn't able to get it all to look right. There are still a few `<li>` elements referenced in `_records-content.scss`

helrond:
OK that should get things working again. One issue that I haven't resolved is related to border-color when you get to the bottom level of a collection. The issue is that the collection styles the border color based on active and hover states, so you get a weird break in the border.

Screen Shot 2020-10-08 at 1 53 03 PM
Screen Shot 2020-10-08 at 1 53 11 PM

I'm not quite sure how to address this - I wonder if we could pad the collection element out so that its borders wrap around the objects? I think ideally we'd want the whole border to either change color on those states, or not change at all.

helrond:
OK I think I got that sorted. Take a look and LMK!

HaSistrunk:
Cool cool. 

This looks good. I unfortunately just learned that aria-controls is on it's way to deprecation, and major screen readers have it turned off by default because apparently it's very verbose and, according to JAWS, "In most cases, the target of the controls relationship is adjacent to the element or does not provide any useful information."

I think this if fine for now, but maybe something to file and iteratively improve down the line. Maybe I'll file an issue in the react-accessible-accordion repo.

HaSistrunk (approved):

helrond:
OK good to know! Definitely something to keep in mind for this project and others...

PR #161, opened 2020-10-09
Adds aria-live message for records detail changes

### LINKED ISSUE #125: Use ARIA live region for collection/object detail changes
When user selects a collection or object in the context section, that change will not be detected unless the user navigates  back to that section on the page.

Implement ARIA live region (polite) on record details so that the screen reader will announce the change in page content. Research how this relates to changing urls (issue #120 ). 

## Additional context
Consider a solution like [react-aria-live](https://www.npmjs.com/package/react-aria-live).

helrond:
Updated: Adds react-aria-live to announce dynamic changes to records details when records are selected in the collection content section.

Adds an aria-live attribute to the `h1` title. 

This works, in the sense that it announces the title whenever it changes, but I'm not sure if this is 100% the desired implementation. Because that title is the same as the button title, you end up hearing something that's like "Important stuff button, important stuff". Thoughts?

fixes #125


HaSistrunk:
Hm, yeah. One problem is that our H1 is not very descriptive of the content (something we may want to think about), so just having the screen reader read it isn't telling the user anything about what has changed. Ideas:
 
1. Implement a visually hidden div similar to the way react-aria-live operates that is updated when a collection/object is selected, and announces something like "details under heading 1 have been updated to describe the selected records" with aria-polite.
2. We could add some helper text on the page, like after the Collection Content heading, that says something like "Select a record to view updated collection details." It may help to have the words "collection details" on the page somewhere, perhaps as part of the H1. They only show up with the mobile toggle right now.
3. Shift focus to the H1 when something is selected. I don't think this is a good option, though, because it messes up the flow if someone is trying to traverse the collection content.
4. Do nothing right now and wait until after usability testing to address/think through other options.

helrond:
OK, I'm inclined to go with option 1, but to do it with react-aria-live. I'll re-request a review when I'm done.

HaSistrunk:
Alright, I think this works pretty well.

PR #163, opened 2020-10-13
Uses children endpoint, change URLs on clicks

### LINKED ISSUE #120: Handle URL changes for clicks within Collections Context

### LINKED ISSUE #147: Use children endpoint to improve performance
There is now a `children` endpoint available in Argo.

Use this endpoint (with pagination) to get faster responses. Will need to be implemented both in RecordsContent and CollectionHitsModal components.

helrond:
Implements `/children` detail endpoint on collections in search results as well as records content components. 

I removed the loading skeleton from RecordsContent because it was pretty jerky and distracting, and the response time is quick enough now that I'm not sure we need it. Would be interested in hearing feedback on that choice.

fixes #147 
fixes #120 

Branched from `issue-125`


helrond:
@HaSistrunk this is ready for review. I've added some code comments which indicate some limitations with `react-accessible-accordion` that negatively impact code complexity. I have filed an issue on that repository, but there doesn't seem to be a ton of activity there so I'm not sure how quickly I'll get a response. 

My proposal would be to review and get this merged, and I'll file another issue which indicates that we need to pay down this particular piece of technical debt, either by rolling our own accordion component or getting this issue resolved in react-accessible-accordion.

HaSistrunk:
Ok, gotcha re: react-accessible-accordion. Yeah, this is fast now! I'm happy to get rid of the loading skeletons if these are the speeds we can reliably expect.

HaSistrunk (approved):

helrond:
I think we can definitely expect these speeds or faster. We can definitely review the loading thing later if we need to though.

PR #165, opened 2020-10-14
Makes MyList modals draggable

### LINKED ISSUE #149: Make modals draggable
## Is your feature request related to a problem?
It would be a nice feature for the user to be able to move the modal on the screen to see the screen underneath.

Consider a component like react-draggable. See this discussion on react-modal's repo: [https://github.com/reactjs/react-modal/issues/232](https://github.com/reactjs/react-modal/issues/232)

helrond:
Uses JQuery and JQuery UI to make modals draggable.

I am OK with leaving this PR open if adding two more dependencies seems out of proportion with the feature value. Happy to discuss!

Fixes #149

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
     "@testing-library/user-event": "^7.1.2",
     "axios": "^0.19.2",
     "formik": "^2.1.5",
+    "jquery": "^3.5.1",
+    "jqueryui": "^1.11.1",
     "node-sass": "^4.14.1",
     "react": "^16.13.1",
     "react-accessible-accordion": "^3.3.3",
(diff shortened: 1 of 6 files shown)


HaSistrunk:
Would [react-draggable](https://www.npmjs.com/package/react-draggable) work instead?

helrond:
@HaSistrunk I was unable to get react-draggable to work. If you read down through that thread there is only one person who was able to get it to work, and in order to do that they had to create a separate content div and hide the one that react-modal uses by default. That seems like a pretty invasive approach which would also likely have some impacts on accessibility. 

My proposal would be to leave this open for now. We have to use jQuery for Universal Viewer anyway, so perhaps once #168 is merged we can revisit this.

PR #166, opened 2020-10-14
Improves interactions on MyList page and associated modals

### LINKED ISSUE #162: Improve my list select and remove pattern
Usability testing showed that users did not clearly understand the conceptual model and functional relationships between the checkboxes and actions including the delete functions.

- Add a confirmation dialog for the Remove All Items button ("Are you sure you want to remove all the items from your list?" with "Cancel" and "Remove" buttons)
- Add a select/deselect all option (this could be another checkbox that aligns with the existing ones)
- Remove the checkboxes on the MyList page, which would shift the selection process to the request modals. This would organize the actions to the My List page (start requests, email, or download), and the selection of which items to include for that specific action to the modal.

- Leave the checkboxes as they are, and take away the Remove buttons. Replace the "Remove all Items" button with a "Remove Selected Items" button that deletes only the items that have been selected using the checkboxes.

helrond:
Makes a number of changes to MyList page and modals:
- Adds a confirm modal when users click on "remove all items" button
- Adds a select/deselect all button on modals
- Removes checkboxes on main list page

fixes #162

### DIFF
--- a/src/__fixtures__/resolvedList.js
+++ b/src/__fixtures__/resolvedList.js
@@ -6,6 +6,7 @@ export const resolvedList = [
       {
         "date": "undated",
         "description": "Architect and business partnter of Wallace Harrison.",
+        "isChecked": false,
         "lastRequested": null,
         "online": true,
         "parent": "Interview Transcripts and Subject Files",
@@ -17,6 +18,7 @@ export const resolvedList = [
       {
         "date": "undated",
         "description": "",
+        "isChecked": false,
         "lastRequested": null,
         "online": undefined,
         "parent": "Interview Transcripts and Subject Files",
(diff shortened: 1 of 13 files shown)


HaSistrunk:
- Can we tie the select/deselect button to the submit button count?
- Can we make the default be that all items are deselected? I'm hear from folks in Reference that it is extremely common for people to accidentally request everything in their list in the old DIMES. Tied with the form validation error message if nothing is selected, I think that we could help solve that issue with an opt-in approach.

HaSistrunk:
So right now whatever is selected persists, right? Like, if you check one item, close the modal, and then open a different modal, that same one item will be checked. I'd be curious to hear your thoughts on that.

helrond:
OK, I've addressed everything you've noted here. There were a couple bugs and removing the checkboxes in the main list does allow us to remove some complexity in the state object, which I've further clarified here.

HaSistrunk (approved):
Yay! This looks good!

PR #176, opened 2020-10-19
Improves styling and behavior of details pane

### LINKED ISSUE #172: Improve styling of details pane
Since we have added extent to the details pane, the display seems off. 

See if the display can be compacted by moving items onto the same line.

helrond:
Makes a number of changes to improve the styling and behavior of the details component:
- compacts the display by showing all short attributes on a single row
- only loads data if it is different than data currently shown
- show line breaks in notes
- add physdesc information to formats

Fixes #172

### DIFF
--- a/src/components/LoadingSkeleton/index.js
+++ b/src/components/LoadingSkeleton/index.js
@@ -81,8 +81,6 @@ export const DetailSkeleton = () => (
     <ul className="panel__list--unstyled">
       <li><Skeleton /></li>
     </ul>
-    <h3 className="panel__heading"><Skeleton /></h3>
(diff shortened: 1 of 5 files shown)


HaSistrunk:
Cool! 

My only reservation is if in the styling, a lot of creators with long names looks too crowded? From my full laptop screen:
image

Or this?
image

I'm happy to give this a shot, though.

helrond:
@HaSistrunk good call - I set `flex-wrap: wrap` and it looks a lot better on these two collections, without making others (that I looked at) look crappy.

HaSistrunk (approved):

PR #192, opened 2020-10-26
Expands ancestors in RecordsContent when loading a URL

### LINKED ISSUE #122: Provide expanded ancestor view at URL referencing an object

helrond:
This is actually a fairly invasive PR, so should be reviewed carefully:
- Replaces `react-accessible-accordion` with local code. Generally speaking, I've used the same patterns (and in some cases cribbed code directly) from this library, but this helps resolve weird issues I was having with loading data into panes after they were opened. This allows us to simplify the RecordsContent components significantly, and get rid of some problematic anti-patterns that would have complicated maintenance. I believe that the accordion is keyboard accessible by WCAG criteria, but a check of that would be good. I've also removed the `aria-controls` attributes per our earlier conversation on #150.
- When a URL is loaded, a `preExpanded` array is constructed from the URIs of the item's ancestors (if any), and any accordion items in that array are expanded. Focus is also applied to the item in the RecordsContent component matching the URL, and it is scrolled into the center of the viewport.
- I've also added a `HitCountBadge` and converted the Component Formerly Known As `HitCount` to `HitCountButton`. I've then implemented the `HitCountBadge` in records context components, since we don't actually want button elements in the list because they're not clickable.
- Last, I did a bit of cleanup in the `PageRecords` component to reduce duplicate code.

fixes #122


HaSistrunk:
Ok great!
One thing I'm not seeing is the heading levels defined in the accordion. We have `role="heading"`, but unless I'm missing it, the `aria-level` attribute for that heading isn't showing up when its rendered (I couldn't figure out why). This does make me wonder, would there be a way to use actual html heading elements instead of `div role="heading"` and `div role="button"` for AccordionItemButton?

Good call with `HitCountButton` and `HitCountBadge`, which also highlights the fact that we have components that look exactly the same, but behave differently in the UI, which isn't the best in terms of predictability/consistency for users. I suspect we'll have ideas about how to solve this after our next round of usability testing, though.

The `preExpanded` array is a great addition. It takes a minute with heavily nested objects with lots of siblings, but I don't see a way around that.

I'd like to test this accordion again more thoroughly as part of a comprehensive site-wide testing with a screen reader in different browsers, but it's looking good to me.

helrond:
OK, I got the aria-level attribute to show up. 

Regarding the aria-level versus HTML header - I was working off of the design pattern here: https://www.w3.org/TR/wai-aria-practices-1.1/#accordion but another reason is that the AccordionHeader component wraps a number of other pieces content, which are themselves HTML elements. All of which is to say that I think a closer look at this with a screen reader would be wise.

PR #202, opened 2020-10-29
Improves performance of MyList

### LINKED ISSUE #199: Improve response time of MyList page load and update
The performance of the MyList page is less than optimal both on load, and also when items are removed from the list.

Find a more performant (or performant-appearing) method of:
- Loading the page.
- Removing items from the list.

helrond:
Makes a number of changes to improve performance of MyList:
- Uses Argo endpoint to return My List data
- Adds methods to remove items from list without reloading
- Slims down props passed from the top level of the app (just the myListCount is passed now)

fixes #199

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -18,38 +18,27 @@ class App extends Component {
     super(props)
     this.state = {
       myListCount: 0,
-      savedList: {}
     }
   }
 
   componentDidMount() {
-    this.setState({savedList: fetchMyList()})
-  }
-
-  componentDidUpdate(nextProps, nextState) {
-    if (nextState.savedList !== this.state.savedList) {
-      this.setState({myListCount: this.countMyList(this.state.savedList)})
-    }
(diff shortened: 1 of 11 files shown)


HaSistrunk:
Yes. So fast now!

From line 16 of `SavedItem/index.js`: 
`{props.parent && <p className="saved-item__found-in">Found in: <a href={props.parentRef}>{props.parent}</a></p>}`

Does that href needs to be updated? It's not showing up in the MyList UI. The link is there, but there is no href attribute.

HaSistrunk:
Also, the first time I open a request modal on the MyList page, the checkboxes are all checked (we want them to be unchecked). If I close and open a new modal without reloading the page, they are then unchecked. So there's a state thing to debug there?

helrond:
@HaSistrunk thanks for the catch on both these things - had made notes to address them and then promptly forgot to before submitting the PR...

HaSistrunk (approved):

PR #203, opened 2020-11-03
Implement SelectInput component

### LINKED ISSUE #201: Align select visual styles and interaction patterns across app
Selects should look and act the same way across the site. They currently do not. The search dropdown has custom styles and uses different HTML elements than the sort results or format dropdowns.

Make sure that selects behave and appear uniformly across the site.

helrond:
Implements a SelectInput component in search bar, search results sort, and duplication modal, which uses downshift's [useSelect](https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect) hook. Styling has been aligned across all instances of the select. In forms, the selected value is supplied via a hidden input.

There are also some styling tweaks to the confirm modals that are part of this PR, and I made sure that the confirm modal is shown when mobile users click on "delete all items".

fixes #201

### DIFF
--- a/src/components/ModalConfirm/index.js
+++ b/src/components/ModalConfirm/index.js
@@ -20,7 +20,7 @@ const ModalConfirm = props => (
     </div>
     <div className="modal-body">
       <div className="modal-message">
-        <p>{props.message}</p>
+        {props.message}
       </div>
     </div>
   </Modal>
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "@testing-library/user-event": "^7.1.2",
     "axios": "^0.19.2",
     "classnames": "^2.2.6",
+    "downshift": "^6.0.6",
     "formik": "^2.1.5",
     "mirador": "^3.0.0-rc.7",
     "node-sass": "^4.14.1",
@@ -24,8 +25,7 @@
     "react-modal": "^3.11.2",
     "react-paginate": "^6.5.0",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "3.4.1",
-    "react-select": "^3.1.0"
+    "react-scripts": "3.4.1"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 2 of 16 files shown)


src/components/ModalConfirm/index.js
```diff
@@ -20,7 +20,7 @@ const ModalConfirm = props => (
     </div>
     <div className="modal-body">
       <div className="modal-message">
-        <p>{props.message}</p>
```
HaSistrunk, line 23: Removing the `<p>` here appears to mess up the text formatting in the modal message (see the Request in Reading Room confirmation).

HaSistrunk:
The SelectInput looks good, and I think is a solid improvement. I'm seeing a couple things with this implementation:

1. I think because we're using the `InputLabel` component for the label, the ids aren't lining up correctly. We should have:<label for="downshift-0-label"...` and `<button id="downshift-0-toggle-button"...`

HaSistrunk:
The SelectInput looks good, and I think is a solid improvement. I'm seeing a couple things with this implementation:

1. I think because we're using the `InputLabel` component props with the label, the ids aren't lining up correctly. I believe we should have:
`<label id="downshift-0-label" for="downshift-0-toggle-button"...>` 
`<button id="downshift-0-toggle-button" aria-labelledby="downshift-0-label downshift-0-toggle-button"...>`
`<ul  id="downshift-0-menu" aria-labelledby="downshift-0-label"...>`

2. When I test with a screenreader, once I select something, it reads "undefined has been selected." I haven't been able to figure out why it won't read the value of the thing being selected. I also tested the example implementation in the downshift documentation, and the sr announces what option has been selected just fine in there.

helrond:
Thanks for this close review @HaSistrunk! I believe I have addressed all of the outstanding issue here, along with a bug in the Duplication Modal that was causing the form not to submit (and a typo which prevented an error message from being displayed)

HaSistrunk:
Looks great!

HaSistrunk (approved):

PR #218, opened 2020-11-11
Makes top navbar sticky

helrond:
Adds a `position: sticky` to make the top nav sticky.

### DIFF
--- a/src/styles/base/_layout.scss
+++ b/src/styles/base/_layout.scss
@@ -9,6 +9,10 @@ form {
   width: 100%;
 }
 
+#main {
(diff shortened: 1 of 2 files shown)


HaSistrunk:
Cool! Apparently `position: sticky` isn't supported on IE or some other older browsers (https://caniuse.com/?search=sticky), but maybe that's ok?

HaSistrunk (approved):

helrond:
I'm not sure it's a huge deal, but it's also not a huge deal to just use `fixed` and add a top margin to `#main` so I did that!

PR #220, opened 2020-11-11
Improve setting of back button URL

helrond:
Thinking about the back button on digital object pages, I realized there were a number of cases in which the current implementation would not work:
1. Users getting to that page via a link from another site
2. Users typing that URL directly into the browser
3. Users getting to that page from the MyList page

I've added a function which *should* account for all of those cases.

I've also added some logic to hide hit count badges if there is no query.

### DIFF
--- a/src/components/PageDigitalObject/index.js
+++ b/src/components/PageDigitalObject/index.js
@@ -72,15 +72,19 @@ const PageDigitalObject = props => {
     ]
   }
 
-  const itemUrl = () => (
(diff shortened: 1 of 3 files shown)


HaSistrunk:
Oh yes, that's a good call.

Wouldn't users typing a URL directly into the browser also be an issue for collection/object urls?

helrond:
Good call - updated the back to search button as well.

HaSistrunk (approved):

PR #226, opened 2020-11-18
Integrates MyList with request broker

helrond:
This is a pretty big PR and deserves some careful review to make sure everything is working as intended.

Removes code used for testing and adds handing for request to Aeon, as well as emails and downloads.

Some additional configuration on RAC systems will likely need to be done in order for emails to work. However, Aeon requests and downloads should work locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
     "downshift": "^6.0.6",
     "formik": "^2.1.5",
     "mirador": "^3.0.0",
-    "moment": "^2.29.1",
     "node-sass": "^4.14.1",
     "react": "^16.13.1",
     "react-accessible-dropdown-menu-hook": "^2.1.1",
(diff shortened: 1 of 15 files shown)


HaSistrunk (approved):

HaSistrunk:
This is looking great and everything seems to be working smoothly for me!

HaSistrunk:
Question: I did add a record that was restricted and see the message about restricted material in the request modal. However, when I add other material that appears to be restricted like objects/QYWj9e74hzsia7Fj9JpHhi, it does not include that message. Is that expected?

helrond:
It's not, but that's also a Request Broker issue, not a DIMES issue. Can you create an issue there with the two URLs and describe the behavior? This restrictions checking thing is going to take some tweaking to get right.

PR #231, opened 2020-11-23
Adds handling for PDF download

### LINKED ISSUE #170: Provide handling for digital object download
Users want to be able to download digital objects.

Add handlers to the download button to support downloading.

## Additional context
Consider this in combination with RockefellerArchiveCenter/pisces#377. Universal Viewer also has (or can have) a download button.

helrond:
Adds handling for PDF download. Although the link has a `download` attribute, the PDF will always open in a new tab. This is because the PDF is hosted on a different domain, and cross-origin downloading is disabled on all modern browsers.

Also adds a new environment variable for this download location.

fixes #170


HaSistrunk (approved):
Great. Just added a title attribute like we did with the fee schedule link to indicate that it opens in a new window

PR #233, opened 2020-11-23
Refactors logic on agent pages

helrond:
Uses `map` instead of `forEach` on agent pages to remove some lines of code. Also hides the "search related collections" if all the related collections can already be seen (i.e. if the search results are less than 8).


HaSistrunk:
Nice! Good call to remove "search related collections" when it doesn't apply.

PR #238, opened 2020-11-24
Display information about downloadable object

### LINKED ISSUE #235: Display size and format of downloadable file
Users need to know what file format (and size) they can expect when they click on the download button.

Add file format and size, as per the design treatments. This will require making a HEAD request against that file.

helrond:
Displays information about a downloadable object, including size and format.

fixes #235

### DIFF
--- a/src/components/Helpers.js
+++ b/src/components/Helpers.js
@@ -35,3 +35,17 @@ export const getFormattedDate = date => {
 
     return month + '/' + day + '/' + year;
 }
+
+
+/** Returns a human readable representation of bytes */
+export const formatBytes = (bytes, decimals = 2) => {
+    if (bytes === 0) return '0 Bytes';
+
+    const k = 1024;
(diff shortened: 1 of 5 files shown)


HaSistrunk:
I think there's an issue with state here, where once the info is displayed for one object (objects/UhixZnaWBmeXGNkZ74hYpE), it displays the exact same info for other objects when I select them. Maybe this is just because that info is not available for these objects, so it doesn't replace what is already there?

helrond:
OK, a couple of things going on here
- requests that errored out were not resetting the `downloadSize` property in state
- using the `props.location.match.id` was not producing the desired result, since that prop didn't update right away. Instead, we now use the identifier from the fetched data.

HaSistrunk (approved):

HaSistrunk:
Ah, ok! Yeah, this is looking good.

PR #239, opened 2020-11-25
Displays selected extent in MyList modals

### LINKED ISSUE #198: Add size of requested materials in MyList
Users should know the quantity of records in their request.

Add a representation of total selected extent to MyList modals.

Further conversation may be required to figure out how do to this:
- Add additional text to the submit buttons
- Add a "total size" sum at the bottom (or the top) of the list
- Other things I'm not thinking of now

helrond:
Curious for feedback about placement and styling of this information...

fixes #198

### DIFF
--- a/src/components/ModalMyList/index.js
+++ b/src/components/ModalMyList/index.js
@@ -94,6 +94,19 @@ const ModalToggleListButton = ({ignoreRestrictions, items, toggleList}) => {
   )
 }
 
+const SelectedTotals = ({ items }) => {
+  const selectedExtents = items.map(
+    g => g.items.map(
+      i => i.submit && i.isChecked ? i.extents : null)).flat(2).filter(i => i !== null)
+  const totals = selectedExtents.reduce((total, current) => (
+    total[current.type] ?
+      {...total, [current.type]: total[current.type] + parseFloat(current.value)} :
+      {...total, [current.type]: parseFloat(current.value)}
+  ), {})
+  const extents = Object.entries(totals).map(e => {console.log(e); return (`${e[1]} ${e[1] === 1 ? e[0] : `${e[0]}s`}`)})
+  return (extents.length ? <p className="selected-totals">{`${extents.join(", ")} selected`}</p> : null)
+}
+
 
 const ModalMyList = props => (
   <Modal
@@ -114,11 +127,13 @@ const ModalMyList = props => (
           ignoreRestrictions={props.ignoreRestrictions}
           items={props.list}
           toggleList={props.toggleList} />
+        <SelectedTotals items={props.list} />
         <ModalSavedItemList
           ignoreRestrictions={props.ignoreRestrictions}
           items={props.list}
           handleChange={props.handleChange}
           setSubmit={props.setSubmit} />
+        <SelectedTotals items={props.list} />
       </div>
       <div className="modal-form">
         {props.form}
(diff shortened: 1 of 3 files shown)


HaSistrunk:
I'm not seeing this showing up... But i don't see any console errors. Could changes with Argo have effected this?

helrond:
I'm definitely seeing it. It's possible that Pisces was unable to parse an extent for any of the items in your list.

HaSistrunk:
Ok, I'll take another look!

HaSistrunk:
Yeah, I think this looks just fine to me styling-wise. The only tiny thing is that it isn't aligned on the right edge with the `modal-saved-items__item-group`, but neither is the button aligned with the items on the left edge, so there's a kind of symmetry there. 

My hesitation is that I had a bunch of stuff in my list to test, and several of them didn't have extent info (Ford microfilm and a couple of AV things with no extent), so the mismatch was a little confusing. I didn't know what would be a folder and what wouldn't. I don't have a handle data-wise on how common this will be, aside from the fact that there is a lot of microfilm. Do we have extents for objects other than folders?

helrond:
So it looks like there are a couple of things happening here, which may impact the usefulness of this feature. Extent calculations, which happen in Pisces, evaluate instances by parsing the `indicator_2` field to determine a value and the `type_2` to get a label for the instance. This generally works for box and folder listings. It does not work for the following scenarios:
- instances have no `type_2` and `indicator_2`, which is the case for many microfilm instances. I'm not sure if there's a useful alternative to display here or not. Is `1 microfilm reel` meaningful or misleading?
- A similar thing is happening with AV items, which have instances but typically no type_2 or indicator_2. Again, I'm unsure what a meaningful alternative would be. Human readable (but not machine-actionable) extents for AV appear to be in physdesc notes.

This may be an idea whose time has not yet come, particularly since we didn't really hear a strong need from users to have this information present. It may also be a case where a partial solution is worse than none ant all (i.e. if inaccurate extent calculations would cause more pain than none).

Thoughts?

HaSistrunk:
The need we did hear from users was for requesting copies - knowing how much they were requesting before agreeing to pay the cost. I could also see value in letting users know that they are requesting microfilm or AV materials, because digitization staff have indicated that folks are sometimes taken by surprise by the format off what they've requested (and I see this in the RR with microfilm). BUT from our previous discussion about copy request formats with Reference, they did not identify this as a major pain point. 

My two ideas:
1. Do not implement this now
2. Follow the lead of old DIMES and keep the folder extent, but then also list the format types for items that are not in folders. That gets to a place where we may want designer input, though.
image

helrond:
OK, that's useful. It sounds to me like there are two problems here:
1. "As a user requesting copies I want to know the original format of the item I'm requesting" - better indication of format of materials in MyList. I agree that's something to add to the design list.
2. "As a user in the reading room, I want to know the size of items independently or in aggregate so that I can plan requests" - improved parsing of extents (which affects both the text that this PR introduces but also extent display in RecordsDetail). I think this is something we may be able to improve by tweaking logic in Pisces (see my earlier comment). I've filed RockefellerArchiveCenter/pisces#407

In the meantime, I think we hold off on merging this PR until we've got better data to work with.

helrond:
@HaSistrunk We should have much better coverage of extents now, so I'd be curious for another review of this PR to see if the functionality is more useful.

HaSistrunk:
Yes, I think so! 
I tweaked it so that `selected:` is at the beginning because I think that reads a bit better when there are a few different formats selected. I also added `selected: 0 items` when there is nothing, because it seems more predictable as a constant presence. Let me know if those changes look ok to you, though.

One (probably rare) thing I see is that when the extent type is "Pages", the logic for singular/plural doesn't work. Eg. objects/Nu5CfRBmTZXEifWKwi5xS9, which has an extent of "82 Pages" which becomes "82 Pagess."

helrond:
Cool, let's merge this and deploy and see where that gets us. I'm not sure how rare the plural thing is, but it strikes me this might crop up in other places in the app. If we start seeing those plural issues I'd argue for us using [pluralize](https://www.npmjs.com/package/pluralize) to handle pluralizing strings.

PR #240, opened 2020-11-25
Improves responsiveness, particularly for large screens

### LINKED ISSUE #236: Home page text is not centered on larger screens
## Describe the bug
The hero on the home page is not centered on large screens.

## To reproduce
Look at the home page on a large screen.

## Expected behavior
The home page should look good on all screen sizes!

### LINKED ISSUE #237: Attributes at the top of the records detail panel need margin/padding
## Describe the bug
The short attributes at the top of the records detail panel can butt into one another.

## To reproduce
Look at collections/H45i6yf7MUHuaRwQVupvg5 on a large screen

## Expected behavior
These attributes should have some margin or padding so they don't run into one another.

helrond:
Makes a number of small changes to CSS to improve site responsiveness:
- Centers the hero on all screen sizes, fixes #236
- Improves styling of the search bar so it is always centered and category select is hidden on mobile.
- Adds a right margin to the short collection/object attributes so they don't run into each other, fixes #237 
- Improves display of download button on mobile.

### DIFF
--- a/src/styles/base/_grid.scss
+++ b/src/styles/base/_grid.scss
@@ -16,7 +16,6 @@ $grid-column: (
 
 .container {
   width: calc(100% - (#{$gutters-default}) * 2);
-  max-width: 1440px;
   margin-left: $gutters-default;
   margin-right: $gutters-default;
   @include grid-container;
--- a/src/styles/base/_typography.scss
+++ b/src/styles/base/_typography.scss
@@ -290,12 +290,10 @@ small {
   line-height: 24px;
   color: $dark-gray;
   @media screen and (min-width: $break-md) {
-    @include grid-column(9);
     font-size: 28px;
     line-height: 38px;
   }
   @media screen and (min-width: $break-lg) {
-    @include grid-column(7);
     font-size: 34px;
     line-height: 42px;
   }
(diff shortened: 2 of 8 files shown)


HaSistrunk:
For large-scale displays bigger than 1440px, which is the max-width of our `container` class, there is some inconsistency in display since we are using both `container` and `container--full-width` (width-100%) in different contexts. So the header, footer, agent page, and my list page have a max width of 1440px, but the search results and collection content/details have a width of 100%.

Maybe instead we can use one consistent `container` class that defines the left and right margins as relative units and gradually increase these values as the screen size gets larger, as [suggested here](https://css-tricks.com/optimizing-large-scale-displays/#containment)?

helrond:
Quick question, on a large screen would we want the records page to look like this:

Screen Shot 2020-12-03 at 12 57 29 PM

helrond:
In the interest of a less-invasive PR for now, I've simply removed the max-width from the `.container` class. 

My thought is that if we want to get into further revisions, it might be wise to have some design treatments created first.

HaSistrunk:
Ok cool. The only thing that jumps out at me with this at a width greater than ~1675 is that on the My List page the `<aside>` buttons line up horizontally instead of stacking vertically. Worth trying to fix now, or revisit with a more thorough review?

helrond:
Ah, good catch. I flexboxed that into submission.

HaSistrunk (approved):

PR #246, opened 2020-12-07
Handles unavailability of request broker

### LINKED ISSUE #243: Solution for when Request Broker is down
When the Request Broker is not functioning, users cannot make a request and there is no indication to users about why they are unable to make a request. They are just blocked from selecting any items.

If the Request Broker is not available, the button instead opens a dialog that tells the user that cannot complete a request.

Proposed error message: Heading: Can't Complete Request Message: "Sorry, our system is unable to process requests right now. We're working to fix this! Please try again later."

We could replace the left side of the existing request modal with an error message. This is more confusing, though. We could also architect the ability to submit requests anyway, but this is a bigger development lift.

helrond:
When MyList page is loaded, pings Request Broker to see if it is available. If the request returns an error code, sets a flag which loads a smaller modal indicating that requesting is out of order. 

I've added RockefellerArchiveCenter/request_broker#70 to implement a more meaningful ping endpoint in that system.

Fixes #243

### DIFF
--- a/src/components/PageMyList/index.js
+++ b/src/components/PageMyList/index.js
@@ -61,6 +61,10 @@
 
     /* Requests CSV data and downloads a local file */
     downloadCsv = () => {
+      if (!this.state.isRequestingAvailable) {
+        this.toggleModal("requestingUnavailable");
+        return;
+      }
       this.setState({ isDownloading: true })
       axios
         .post(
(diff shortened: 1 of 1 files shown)


HaSistrunk:
Should we have the same out of order modal pop up for downloadCsv and emailList?

helrond:
Oh! Yes, good call - I was very focused on Aeon but those use the Request Broker endpoints too...

HaSistrunk (approved):

PR #247, opened 2020-12-07
Adds loading overlay for Record Content

### LINKED ISSUE #245: Add loading indication when opening an object from its url
From usability testing, we observed that the time it takes to load/scroll to a specific object (as in opening to a specific object's url) can be disorienting or the user does not understand that this is happening.

Add an overlay indicating that the page is loading until it scrolls. Accessibility-wise, focus should be set to the element containing the loading message (if just an icon, give it an aria-label with a loading message), and once it is loaded, transfer focus to the object that loaded.

helrond:
Adds a loading overlay for RecordsContent component which appears when records are loaded from a URL.

@HaSistrunk the styling could definitely use some tweakage, so feel free to mess with that.

fixes #245


helrond:
@HaSistrunk it looks like that merge might have reverted the query highlight stuff I did yesterday.

HaSistrunk:
> @HaSistrunk it looks like that merge might have reverted the query highlight stuff I did yesterday.

Oh shoot, sorry. I'll revert.

HaSistrunk:
I just tweaked a few things and added some styles for small screens. They should work with the recent changes to improve responsiveness for small screen that are in `base`, and that I am now afraid to merge here.

Do you think we need to keep the overlay from covering the footer, or nah? I was going to do that, but then it got a little tricky with the different screen sizes.

HaSistrunk (approved):

helrond:
I'm not super worried about the overlay covering the footer - as content loads it will push that footer down and hide the footer anyway.

Issue #248, opened 2020-12-07
Create Mirador plugin for item details button


helrond:
## Is your feature request related to a problem? Please describe.
The item details button is currently a separate component which is statically positioned on top of the Mirador header. This makes it hard to style in a responsive way and mesh with underlying Mirador styles.

## Describe the solution you'd like
Add a Mirador plugin for this button so that the header can be styled as a whole.

## Describe alternatives you've considered
It may be possible to add additional styles to Mirador without going the route of a plugin. Other solutions may also be available.

PR #249, opened 2020-12-07
Improve responsiveness, particularly on smaller screens

### LINKED ISSUE #244: Improve responsiveness
There are a few layout and element size issues on mobile to button up:

1. Width of less than ~370 px (small mobile phone like iphone SE). The "filter" and "sort by" menus stack and need to do so with some space and be aligned:
image

2. At ~375px width (iphone X), the filter modal close button and longer filter text start to be cut off:
image
image

3. On med and small screens the collection content/detail layouts aren't displaying as intended. The content runs over and below the footer if you keep scrolling down from the details. This needs a little investigation to figure out what is going wrong. Possibly related: the nav hamburger menu button on this page is very far to the right outside of the display.

4. In digital object viewer on mobile, "Back to Item Details" link overlaps with object title:
image

5. In My List request modals, reCAPTCHA runs off the screen before the break point at ~1025px.

6. In My List request modals, when the submit/cancel button stack, they should have a margin between them.
image

7. A slightly unrelated extra: On the My List page, there is more space between the navigation link ("start a new search") and the h1 than on other pages. Can we tighten that up for consistency?

helrond:
Improves responsiveness of various pages and components:
- fix stacking and indent of filter and sort by components on search page for small screens
- add max-width to facet modal to fix text overflow
- apply `.hidden` styles to records content on small screens so so components aren't visible in footer.
- prevent reading room request button from wrapping in MyList sidebar
- Improve button stacking in MyList modals
- improve captcha sizing breakpoints
- Ensure form background always goes to bottom of MyList modals
- remove top padding from MyList header
- adds media query mixins and replace existing media queries

fixes #244

### DIFF
--- a/src/components/ModalConfirm/index.js
+++ b/src/components/ModalConfirm/index.js
@@ -18,7 +18,7 @@ const ModalConfirm = props => (
         <MaterialIcon icon="close"/>
       </button>
     </div>
-    <div className="modal-body">
+    <div className="modal-body--confirm">
       <div className="modal-message">
         {props.message}
       </div>
--- a/src/components/ModalSearch/index.js
+++ b/src/components/ModalSearch/index.js
@@ -37,7 +37,7 @@ export const FacetModal = props => {
           <MaterialIcon icon="close"/>
         </button>
       </div>
-      <div className="modal-body">
+      <div className="modal-body--search">
         <Facet title="View Online">
           <CheckBoxInput
             id="online"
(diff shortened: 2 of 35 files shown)


HaSistrunk:
Awesome! Just a couple things I see:

1. Can you look again at the facets modal? It looks great on mobile, but not so much on a larger screen:

image

2. The other thing I noticed is those pesky My List modals, which are now spreading their wings.

image

helrond:
Good catches - I think I got a handle on both of these now!

HaSistrunk:
In a final check I was still getting some wonkiness with the captcha on phone-size mobile screens, so I made some changes there. Let me know if I've done something I shouldn't to those styles.

And just a note to say that I see fixing the item details button in Mirador is tied to #248.

HaSistrunk (approved):

PR #256, opened 2020-12-07
Only highlight queries on components with hit counts

### LINKED ISSUE #251: Improve query term highlighting
Query term highlighting is not tied to actual hits, and can be misleading.

Find a way to more effectively tie query highlighting to hit counts. Will also need to add query highlighting to term text, since that is now being searched.

helrond:
This is not totally perfect, but it's A LOT better.

Queries will only be highlighted on components that also have a hit count. While this will still result in some spurious query highlighting, it removes the really misleading highlights.

fixes #251

### DIFF
--- a/src/components/QueryHighlighter/index.js
+++ b/src/components/QueryHighlighter/index.js
@@ -4,8 +4,8 @@ import Highlighter from "react-highlight-words";
 import "./styles.scss";
 
 
-const QueryHighlighter = ({ hitCount, query, text }) => {
-  const parsedQuery = query && hitCount ? query.replace(/"([^"]+(?="))"/g, '$1').split(" ") : []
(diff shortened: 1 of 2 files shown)


HaSistrunk:
Yeah, this is a nice change. Since it's tied to hit count, that means that there are no longer highlights in the description text in the record details, right? Is there a way to retain that description notes highlighting with this change?

helrond:
OK, I took a slightly different approach which will always show query highlighting in RecordsDetail text (since we have no hit count available to us there). I *think* that's an okay approach because you'll only see that if you click on a component. In the RecordsContent component, query highlighting still only happens on components with hit counts.

HaSistrunk:
Great, let's try it! I feel like that should hopefully be a good balance.

HaSistrunk (approved):

PR #267, opened 2020-12-11
Improves toggle button accessibility

### LINKED ISSUE #262: Improve button accessibility
When using VoiceOver in Chrome and Safari:

1. Buttons that have label text that changes when pressed do not announce this change to the screenreader.
- Select all items / Deselect all items
- Add to list / Remove from list
- Download as .CSV / Downloading (this is a little different, and is fine as is in that it will be announced when button is pressed, button is disabled when downloading, and the SR announces when the download is complete).

2. "Add to list" button text is read as "A D D to list" by SR. 

## Expected behavior
1. These are basically toggle buttons, so the user should be able to interact with them in a way that indicates what pressing the button will do while they are focused on it even if they have just pressed it without having to leave and then refocus.

2. "Add to list" button should be read correctly.

1. Give each button a single aria-label that does not change, like "Add to list" and "Select all items," and then add an aria-pressed attribute to the button to indicate if it is pressed or not. This would indicate to the user that what the label indicates is either off or on. If there is an aria-label present, the SR will not read the button label. 
Downsides: Not be as clear as two descriptive labels, and can be confusing for sighted SR users.

- Make the "Select all items" button a checkbox instead of a button with a label that reads "Select/deselect all items."

- Do something with state that I don't fully understand. See the recommendations in [this post](https://sarahmhigley.com/writing/playing-with-state/).

helrond:
Adds aria-label and aria-pressed attributes to toggle buttons:
- Select/Deselect items (MyList modals)
- Add to/Remove from list (RecordsContent)
- Show all/show less items (Search facet)

I did not change the download button since your comment seemed to indicate it was okay, but I can easily do that.

fixes #262

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -13,6 +13,7 @@ const Button = (props) => (
     aria-label={props.ariaLabel}
     aria-haspopup={props.ariaHasPopup}
     aria-expanded={props.ariaExpanded}
+    aria-pressed={props.ariaPressed}
     disabled={props.disabled} >
       {props.iconBefore && <MaterialIcon icon={props.iconBefore} />} {props.label} {props.iconAfter && <MaterialIcon icon={props.iconAfter} />}
   </button>)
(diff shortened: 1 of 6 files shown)


HaSistrunk (approved):
Great! I just tweaked the aria-label for "Toggle select all items" because the screenreader announces "Toggle select all items, toggle button", and "Selected, Toggle select all items, toggle button". Cutting the redundancy.

PR #268, opened 2020-12-11
Improves accessibility of truncated checkboxes in facets

### LINKED ISSUE #263: Improve accessibility of truncated checkboxes in facets
In the facets modal, the checkboxes that are revealed using the show all/show less button are hidden using css, so they are not hidden from screenreaders or keyboard focus. There are two problems with this:

- The show all/ show less button doesn't have any function for users using SRs since they can access all the content without it.
- Sighted users using the keyboard or other AT to navigate won't be able to see what element has focus when they are accessing the visually hidden inputs in the tab order.

Not sure the best approach for how to make this happen, but the facet items should be programmatically hidden until the "show all" button is pressed. `overflow:hidden` only hides things visually.

The `disclosure` element could be used, but it doesn't exactly fit our current design.

## Additional context
See #262 for issue related to button text that changes when the button is pressed.

helrond:
Slices the list of facet terms, which adds/removes those items from the DOM, rather than showing/hiding with CSS. Also tweaks the facet input styling for better display of long values that wrap.

Fixes #263

### DIFF
--- a/src/components/Facet/index.js
+++ b/src/components/Facet/index.js
@@ -32,7 +32,7 @@ const Facet = ({ children, handleChange, items, paramKey, params, title }) => {
   const toggleOpen = isOpen => {
     setIsOpen(!isOpen)
   }
-  const facetValues = items ?? [];
+  const facetValues = items ? isOpen ? items : items.slice(0,5) : [];
   const isChecked = key => {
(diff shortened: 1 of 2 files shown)


HaSistrunk (approved):
This works well!

PR #270, opened 2020-12-11
Improves understanding of link purpose

### LINKED ISSUE #266: Improve understanding of link purpose
- The Schedule a Visit button doesn't indicate that it will open an email, which should be better clarified particularly for screenreader users.
- In footer, "By appointment only" link text is not descriptive of its target

- Change the “Schedule a visit” button to be a link with `title=“opens email”`. Alternatively, if we want to keep this as a button, we can add an aria-label that gives a better indication of what will happen if the user presses it - "Email to schedule a visit."
- In Footer, either change "By appointment only" link text or add an aria-label to the link that reads something like "How to access and request materials."

## Additional context
See [WCAG SC 2.4.4. - Link Purpose (In Context)](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)

helrond:
- Changes the `Schedule a Visit` button (on med-up) to a link, adds a `title` to these links on all screen sizes.
- Updates the text in a footer to be more descriptive of content at the other end of the href.

fixes #266

### DIFF
--- a/src/components/Dropdown/index.js
+++ b/src/components/Dropdown/index.js
@@ -23,6 +23,7 @@ const Dropdown = (props) => {
       className={classnames("btn", itemClassName)}
       onClick={item.handleClick}
       href={item.href}
+      title={item.title}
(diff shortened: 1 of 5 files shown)


HaSistrunk (approved):
Great!

PR #273, opened 2020-12-14
Truncates description text

### LINKED ISSUE #264: Hide truncated description text in collection content
The item description text for collections and their children in the RecordsContent component is not hidden from SRs, so lengthy text is very verbose, redundant, and slows navigation.

Hide the truncated text from all users. The current `text--truncate` class uses `overflow:hidden`, which only visually hides the content.

helrond:
This was surprisingly non-trivial to figure out. There are a number of libraries out there which will trim a string to the size of its parent container, but I was not a huge fan of introducing another dependency just for a single use. Those also created additional complexities for hit highlighting.

So, what I've done here is a bit of a compromise. I've truncated text to a set character length which is long enough that it just overlaps large screens. I've also kept the CSS styles in place. What this means is that screen readers will generally read a slightly longer string than is visible.

Fixes #264

### DIFF
--- a/src/components/Helpers.js
+++ b/src/components/Helpers.js
@@ -49,3 +49,15 @@ export const formatBytes = (bytes, decimals = 2) => {
 
     return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i];
 }
+
+/** Trims a string to a specified length */
+export const truncateString = (text, maxLength) => {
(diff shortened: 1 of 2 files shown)


HaSistrunk:
Ok great, I think this is a nice solution and majorly improves the SR UX.

HaSistrunk (approved):

Issue #66, opened 2020-08-18
Improve focus styles


HaSistrunk:
## Is your feature request related to a problem? Please describe.
We don't have specific focus styles on all the focusable elements from the designs, but a consistent approach to custom styles would be an improvement on the browser default styles.

## Describe the solution you'd like
Ensure all interactive elements have clearly visible focus styles, and customize styles as necessary when browser defaults are insufficient

PR #274, opened 2020-12-14
Adds visual indication of focus state to checkboxes

### LINKED ISSUE #265: Visible checkboxes don't receive keyboard focus
## Describe the bug
I missed this before, but the way we've implemented the checkboxes in the facets and saved items in the modals, they do not show keyboard focus because the input us hidden an only the label is visible. 

## To reproduce
1. Open a request modal
2. Use tab to move focus through the elements
3. See that the checkboxes do not receive focus. 

## Expected behavior
The checkboxes should have a visual indication of focus.

helrond:
Adds initial focus styles for checkboxes. This will need to be improved, but maybe that can be part of #66?

Fixes #265

### DIFF
--- a/src/styles/components/_inputs.scss
+++ b/src/styles/components/_inputs.scss
@@ -73,7 +73,6 @@ label {
   top: 0px;
   width: 16px;
   height: 16px;
-  margin-left: 3px;
 }
 
 /* checked mark aspect */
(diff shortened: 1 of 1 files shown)


HaSistrunk:
Yes, this looks good! Made some little tweaks, but will look more comprehensively as part of #66 .

HaSistrunk (approved):

HaSistrunk (changes requested):
Actually, wait. I see that the focus outline is getting cut off in the filter checkboxes. Let me take a closer look.

HaSistrunk (approved):

PR #278, opened 2020-12-15
Improve and standardize focus styles

### LINKED ISSUE #66: Improve focus styles

HaSistrunk:
Standardizes button and link focus styles across browsers, adds focus styles where there were none, and makes the focus indicators a little more strident and easier to see. I tried to balance the aesthetics so things aren't too distracting for mouse users, but I'd love a tab-through to check that these look ok (including the accordion buttons in record details).

I didn't mess too much with the browser default styles for inputs.

Fixes #66

### DIFF
--- a/src/components/RecordsDetail/index.js
+++ b/src/components/RecordsDetail/index.js
@@ -181,7 +181,7 @@ const RecordsDetail = props => {
     }
     <Accordion className="accordion accordion--details" preExpanded={["summary"]} allowZeroExpanded={true}>
       <AccordionItem className="accordion__item" uuid="summary">
-        <AccordionItemHeading className="accordion__heading" ariaLevel={2}>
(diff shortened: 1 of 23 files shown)


helrond (approved):
Overall I think this is fine for now. I tweaked the homepage styles a bit because the outline focus wasn't really visible on the dropdown or the search button.

Is this something we want to come back to later with some design input?

HaSistrunk:
> Is this something we want to come back to later with some design input?

I think it would at least be worth getting a review from someone with a design mind, as long as we communicate the accessibility considerations.

PR #280, opened 2020-12-16
Improve viewer styles so nav displays

### LINKED ISSUE #279: Nav header doesn't appear in digital object viewer
## Describe the bug
Nav header isn't showing up in the digital object viewer, although it is in the DOM, so this appears to be a layout issue. I'm wondering if PR #218 's solution to use `fixed` and add a top margin to #main effected this based on what is designated as #main on this page?

## To reproduce
Open the digital object viewer. Example: /objects/UhixZnaWBmeXGNkZ74hYpE/view. Notice that you can inspect the page and see the header in the code, as well as tab through the links, but it is not visible.

helrond:
fixes #279

### DIFF
--- a/src/styles/base/_utilities.scss
+++ b/src/styles/base/_utilities.scss
@@ -15,7 +15,7 @@ $transition-default: all 0.2s ease;
       overflow: auto;
       position: absolute;
       width: auto;
-      z-index: 99;
+      z-index: 1000;
   }
 }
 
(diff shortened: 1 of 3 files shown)


HaSistrunk:
The only problem I see here is that the skip link and nav menu in mobile drop down below that top 60px, so they don't show up under the viewer. If this is too obnoxious to implement, we can leave the nav header out. However, I identified this issue when testing the keyboard focus styles for issue #66, because the header still receives keyboard focus even when you can't see it. So if it is easier to leave off altogether, perhaps we can at least make sure the header is entirely hidden (not just visually covered up?).

helrond:
Good catch! That should be fixed now...

HaSistrunk (approved):

PR #283, opened 2020-12-21
Correctly pluralize words

helrond:
Uses the `pluralize` library to pluralize extents in MyList modals and RecordDetail components.

Also includes a tweak to the string length calculation for cases where the string is very close to the max length.


HaSistrunk:
This is cool. I see in the records detail that it does handle "cubic feet" as "cubic feets" though...

HaSistrunk (approved):

PR #287, opened 2020-12-22
Implement gtm for google analytics tracking

### LINKED ISSUE #285: Implement Google Tag Manager
Add the gtm snippet to enable analytics

HaSistrunk:
Adds Google Tag Manager snippet to `index.html`.

Question: is there a need to keep this code in Javascript files? If so, [react-gtm-module](https://www.npmjs.com/package/react-gtm-module) may be a better solution, but I wasn't sure that was necessary.

Note: This leaves the handling of pageview data when there is no new page load to Tag Manager (using the History Change trigger).

Fixes #285

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -24,8 +24,19 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>DIMES: Online Collections and Catalog of Rockefeller Archive Center</title>
+    <!-- Google Tag Manager -->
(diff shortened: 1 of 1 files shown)


helrond:
@HaSistrunk I don't know of a reason we'd need to put this in a JS file, and it honestly seems unnecessary to add another dependency just to add some stuff to `<head>`. I don't know a lot about Tag Manager, but this looks good from an eye test perspective.

helrond (approved):

PR #306, opened 2021-01-15
Hides download and view buttons for digital assets when they are missing

### LINKED ISSUE #292: Add logic to account for components with DAO links but no digitized files
Since we're likely still to be in the process of converting images on February 1st (and in order to provide as much flexibility as possible in automating image derivative creation processes) we should implement logic which looks for digitized files in addition to checking for a DAO link when determining whether or not an object is online.

Add additional checking which checks for the presence of a digitized file (either by looking for a PDF, or by looking for the IIIF manifest) on page load in addition to checking the "online" key.

## Additional context
There are a number of places where this will have UI/functional implications:
- View online/download buttons in RecordsDetail
- View online buttons in RecordContent
- View online buttons in MyList
- MyList parsing (some of this is tied to the Request Broker, so may be more complicated/less desirable to change)

helrond:
In order to cover cases where a digital asset has not yet been uploaded for a digital object, added logic to hide download and view buttons when a `HEAD` request against the PDF asset fails.

This was done in the RecordsDetail, RecordsContent and MyList components/pages

fixes #292

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,4 @@
 REACT_APP_ARGO_BASEURL=http://api.rockarch.org
 REACT_APP_REQUEST_BROKER_BASEURL=http://rac-vch.ad.rockarchive.org:8001/api
 REACT_APP_LOCALSTORAGE_KEY=DIMESMyList
+REACT_APP_S3_BASEURL=https://iiif.rockarch.org
--- a/src/components/PageRecords/index.js
+++ b/src/components/PageRecords/index.js
@@ -58,7 +58,6 @@ class PageRecords extends Component {
               this.setState({ downloadSize: formatBytes(res.headers["content-length"]) })
             })
             .catch(e => {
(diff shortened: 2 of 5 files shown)


HaSistrunk:
I'm thinking these objects also should not show up as results when using the "show me digital materials only" filter?

helrond:
Ideally yes, but that's much harder because it would mean checking at the time of indexing whether or not a derivative exists. We'd also need a way to go back and repeatedly check for items which have a DAO link but are missing a digital object. My hope is that the number of objects which have a DAO link but no asset is pretty small at any given time. We can always revisit this later if this turns out to be a bigger problem, but it would mean figuring out how to do this in the data pipeline.

See RockefellerArchiveCenter/pisces#428 and RockefellerArchiveCenter/pisces#429

HaSistrunk (approved):

PR #307, opened 2021-01-15
Updates to enable Google Tag Manager implementation

HaSistrunk:
A few small tweaks for better implementation of event tracking in GTM:
- Adds ids to header links and facet titles
- Adjusts pointer events in css to prevent icons and svgs from being processed as click events


helrond (approved):
Very nice!

PR #318, opened 2021-01-25
Correctly hide download buttons when no download is available

### LINKED ISSUE #315: View online button persists when it should not
## Describe the bug
The "view online" button in item details does not go away when the user selects an object that is not available to view online after selecting one that is.

## To reproduce
1. Go to objects/NTgrrvWPyMzgPyL4UU7Spb
2. Click on the object just below the first object (Atomic energy - General)
3. Notice that the "View Online' button in the object details is still visible

## Expected behavior
Button should disappear when digital object is not available for a given object

helrond:
Explicitly set downloadSize in component render. Fixes #315


HaSistrunk (approved):
Cool, that works!

PR #321, opened 2021-01-26
Sends a custom pageview event

helrond:
Sends a custom event `reactPageViewEvent` after page title has been correctly set.


HaSistrunk (approved):
Great, this works well along with a custom reactPageViewEvent Google Tag Manager event to trigger a page view tag. Everything seems to functioning nicely.

PR #343, opened 2021-02-05
Highlights entire query phrase

### LINKED ISSUE #340: Improve hit highlighting for phrases
Now that we have changed the default behavior for DIMES to use boolean AND we should update hit highlighting so it only highlights matching phrases.

Remove behaviors that highlight each individual word in a search phrase.

helrond:
Removes `split` on spaces, so entire query phrase is highlighted.

Fixes #340

### DIFF
--- a/src/components/QueryHighlighter/__tests__/QueryHighligher.test.js
+++ b/src/components/QueryHighlighter/__tests__/QueryHighligher.test.js
@@ -15,7 +15,7 @@ afterEach(() => {
   container = null;
 });
 
-it('renders props correctly', () => {
+it('finds word at beginning', () => {
   act(() => {
     render(<QueryHighlighter query="foo" text="foo bar baz" />, container);
   })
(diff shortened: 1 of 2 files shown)


HaSistrunk:
I want to acknowledge the tradeoff in that an AND search doesn't have to be a phrase. So a search for Nelson Rockefeller won't highlight anything in the title/description if they read Nelson A. Rockefeller, for example. Or any two words that may be in the description, but don't appear together. Those would be a hit, but no terms highlighted.

helrond:
Yes, that's true. I think this comes down to whether it's worse to highlight some words that aren't hits, or to not highlight some words that are. 

I'm also noting that `react-highlight-words` allows you to write your own `findChunks` method, so we could potentially use that to introduce more Boolean-like logic.

HaSistrunk:
Ok, good find. I think `react-highlight-words` is a good in-between solution in that it will align pretty closely with our default Boolean 'AND' search.

HaSistrunk (approved):

PR #344, opened 2021-02-05
Using Helmet callback to send page view event

### LINKED ISSUE #339: Page titles not consistently sent to analytics
The page title is not consistently set before page title data is available to Tag Manager, and results in a `(not set)` page title that defaults to the "DIMES: Online Collections and Catalog of Rockefeller Archive Center" that is in `index.html`.

This occurs inconsistently, but apparently on all pages. Google Analytics data from Feb 1-3 shows that of 1050 pageviews with the "DIMES: Online Collections..." page title, 365 were from URLs other than the actual home page.

The page title associated with the current page would be correctly sent to Tag Manager.

## Additional context
Things we've tried:
- Created custom pageview event (PR #321) to send pageview data to Tag Manager. This solved a problem in which page title info was delayed in being sent until the next page was loaded, but not the issue of no page title being sent.
- Tried using an older version of React Helmet (did not fix).
- Tried implementing Google Tag Manager via the [react-gtm-module](https://www.npmjs.com/package/react-gtm-module) instead of putting the gtm snippets directly in `index.html` (did not fix).
- Tried wrapping the `firePageViewEvent` helper in a `setTimeOut()` function with various ms times specified (did not fix).

helrond:
This revises the `firePageViewEvent` to be triggered by the Helmet `onChangeClientState` callback. Additional code has been added so the page view event is only fired once.

Fixes #339

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+cache: yarn
 sudo: required
 env:
   global:
--- a/src/components/Helpers.js
+++ b/src/components/Helpers.js
@@ -71,11 +71,21 @@ export const truncateString = (text, maxLength) => {
 
 /** Sends a custom pageview event to Googgle Tag Manager.
 * This allows us to ensure that the correct page titles are sent.  */
-export const firePageViewEvent = () => {
+var done = false;
+var prevTitle = ""
+export const firePageViewEvent = title => {
+  if (title && title !== prevTitle) {
+    done = false;
+  }
+  if (title && !done) {
     if (window && window.dataLayer) {
-        let dataLayer = window.dataLayer || [];
-        dataLayer.push({
-            'event': 'reactPageViewEvent'
-        });
+      let dataLayer = window.dataLayer || [];
+      console.log("push", title)
+      dataLayer.push({
+          'event': 'reactPageViewEvent'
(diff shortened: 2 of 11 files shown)


HaSistrunk:
@helrond, I moved the page view event for all pages to Helmet. I also removed the "digital item - " piece of the Viewer page title because the page view event was firing twice for that title and this solved it. I know we added that for a more descriptive page title that distinguishes the viewer from the records, but it seemed like the simplest way to solve the issue.

Let me know if this still looks ok! I've tested pretty rigorously and things seem to be functioning as expected in Tag Manager/Analytics with this solution.

HaSistrunk:
@helrond, I tweaked a few things if you could confirm that this looks ok:
- I moved the page view event triggers for all pages to Helmet. 
- I also removed the "digital item - " prefix of PageDigitalObject because the page view event was firing twice there and that was the simplest way to fix it. I know we added that more verbose title for clarity to distinguish between the record page and the viewer, so I welcome any better ideas there, but I'm ok with it as is.

helrond:
Cool - I just tweaked the title in PageDigitalObject because the string interpolation wasn't necessary. I've also added a key to `.travis.yml` to cache build dependencies.

PR #351, opened 2021-02-09
Remove HEAD requests and associated code

### LINKED ISSUE #348: Remove HEAD requests in RecordsContent and MyList
Once RockefellerArchiveCenter/pisces#431 is merged, we can rely on the `online` boolean as a solid indicator of whether or not an item has an online asset.

Remove the `HEAD` requests made simply to check if an online asset exists. We'll need to keep the `HEAD` request in the RecordsDetail component, since that is how the size of the PDF is set.

helrond:
Removes `HEAD` requests which confirm that an online asset exists, as we have now moved that functionality to the data pipeline. Fixes #348

### DIFF
--- a/src/components/RecordsContent/index.js
+++ b/src/components/RecordsContent/index.js
@@ -24,7 +24,6 @@ export class RecordsChild extends Component {
     this.state = {
       children: [],
       isSaved: false,
-      hasOnlineAsset: false,
     }
   }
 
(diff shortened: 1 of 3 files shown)


HaSistrunk:
Ok cool. And no need to update any tests for this change, right?

HaSistrunk (approved):

helrond:
Nope - this is all in Page components, which I haven't written tests for yet.

PR #356, opened 2021-02-11
Displays Biographical/Historical notes

### LINKED ISSUE #355: Add bio/hist note to RecordsDetail
Because the bio/hist note is indexed, there are search term matches contained in these notes that users can't see in the UI.

Add the bio/hist notes to display in RecordsDetail. 
We'll also want to consider implementing a matches badge in the collection details for collections that have search matches in their content, because currently the matches badges only reflect matches in objects. Perhaps a good place for this is after the "Found in" section and before the Description? 

Note that these can be very long notes, so will often add a lot of text to the collection details.

Could also not index these notes so that there are no search matches. However, it's probably better to show instead of hide info from users where it exists.

## Additional context
Overall, this is something we should consider to include in future usability testing.

helrond:
Fixes #355

### DIFF
--- a/src/components/RecordsDetail/index.js
+++ b/src/components/RecordsDetail/index.js
@@ -213,7 +213,7 @@ const RecordsDetail = props => {
                 text={props.item.description} />
               <PanelTextSection
                 params={props.params}
-                title='Description'
+                title='Biographical/Historical Note'
                 text={noteTextByType(props.item.notes, 'bioghist')} />
               { props.item.notes && props.item.notes.filter(n => n.type === 'odd').map(n => (
                 <PanelTextSection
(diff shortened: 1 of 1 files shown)


src/components/RecordsDetail/index.js
```diff
@@ -211,6 +211,10 @@ const RecordsDetail = props => {
                 params={props.params}
                 title='Description'
                 text={props.item.description} />
+              <PanelTextSection
+                params={props.params}
+                title='Description'
+                text={noteTextByType(props.item.notes, 'bioghist')} />
```
HaSistrunk:
```suggestion
              <PanelTextSection
                params={props.params}
                title='Biographical/Historical Note'
                text={noteTextByType(props.item.notes, 'bioghist')} />
```

HaSistrunk: Looking to avoid double description headings: image

helrond: 🤦 Good call - got a little too trigger happy on the cut and paste there...

HaSistrunk (approved):

PR #364, opened 2021-03-26
Displays suggested search terms 

### LINKED ISSUE #272: Display suggested search terms
Misspellings of names or other terms can cause users to miss hits.

Display suggested search terms. Would need to decide if these suggestions would always be display, or only sometimes. If it's the latter, we'd then need to come up with some criteria for when they are displayed.

## Additional context
We should probably wait for a design solution before trying to implement this.

helrond:
When search terms have zero results, displays a page which indicates that no results have been found, and provides some linked search term suggestions.

This will only work when run against the dev Argo instance.

Fixes #272

### DIFF
--- a/src/components/PageSearch/index.js
+++ b/src/components/PageSearch/index.js
@@ -29,6 +29,7 @@ class PageSearch extends Component {
       resultsCount: 0,
       facetIsOpen: false,
       facetData: {},
+      suggestions: []
     };
   };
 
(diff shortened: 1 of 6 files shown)


HaSistrunk:
This looks good! I also looked at the Argo changes to add the suggest filter. Am I understanding correctly that it looks at each term in the search query separately and returns suggested terms for each term that doesn't return hits?

helrond:
That's right - I was not getting very useful results for phrases, so I went with terms instead. Unless you disagree, my thought was that we should test this and see what users have to say about the suggestions. Getting quality phrase suggestions back will likely require additional work in Elasticsearch.

HaSistrunk (approved):

PR #365, opened 2021-03-29
Separate type dropdown from search input and group with digital object checkbox

### LINKED ISSUE #217: Improve awareness of and access to digital materials
Better surface the fact that there is digital content available to view online and facilitate users' ability to find that content.

Add "digital content" or "digital materials" to the dropdown option in the search select menu. Using "digital content" as a descriptor mirrors the language used on the homepage: "Search Our Collections. Discover People and Organizations. Access Digital Content," but our current facet read "show me digital materials only." I'm not sure which is better.

1. Add a checkbox associated with the search form that specifies something like "show me digital materials only," as a way to include that facet at the start of the search process.
2. Do nothing and keep testing this with users.

helrond:
Removes the digital object checkbox from the "Filter Search" modal, and groups it with the type dropdown.

The styling felt a little messy on this one as I was doing it - would appreciate a second set of eyes on that in particular.

Fixes #217

### DIFF
--- a/src/styles/components/_search-form.scss
+++ b/src/styles/components/_search-form.scss
@@ -11,6 +11,7 @@ input[type=search] {
   display: flex;
   justify-content: center;
   flex-wrap: wrap;
+  max-width: 1300px;
   margin: 30px 30px 110px 30px;
   @include md-up {
     margin: 59px 75px 190px 75px;
@@ -34,7 +35,9 @@ input[type=search] {
 }
 
 .search-form--results {
-  max-width: 700px;
+  display: flex;
+  flex-wrap: wrap;
+  width: 100%;
 }
 
 .input-group__search {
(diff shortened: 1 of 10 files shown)


HaSistrunk:
Awesome! I futzed with the styles some - take a look and see what you think. I added the blue background color to the checkbox select like the design have, but I do wonder if this indicates a toggle button which is misleading. Is that why you didn't implement initially (happy to reverse).

A couple responsiveness things:
1. The search button on the homepage has the "Search" text now, but I'm wondering if that's too wide on mobile? It doesn't leave a lot of space for the text input on small screens.
2. Can you take a look at the responsiveness on the search results page for the search form? It needs a margin between the search input and search controls at ~1700px width, but I was having trouble solving that rubiks cube.

helrond:
Great, thanks for tweaking. I feel like it takes me a couple of days to get back into Sass after I've been away for a while.

See how you feel about the latest tweaks. I think the search results page is looking better now, and I did some tweaking so the `Search` text won't show up on screens smaller than 580px. I re-added the aria label since there's not always text.

I think I'm okay with the digital materials checkbox looking like a toggle. Let's test and see how users respond.

src/styles/components/_search-form.scss
```diff
@@ -56,23 +94,28 @@ input[type=search] {
...
+.select__search__wrapper {
+  display: none;
+  margin-right: 14px;
+  @include md-up {
+    display: block;
+  }
+}
+
 .select__search__control {
   display: none;
```
HaSistrunk, line 110:
```suggestion
```

HaSistrunk:
I'm still seeing the 'Search' text on mobile, but I'm not sure why because your code makes sense to me.

helrond:
Depending on how you're looking at mobile, you'll likely need to reload the page to get that change to happen.

HaSistrunk:
> Depending on how you're looking at mobile, you'll likely need to reload the page to get that change to happen.

Oh of course, thanks. Looking good!

PR #366, opened 2021-03-30
Scroll focused element into view, improve focus styles

### LINKED ISSUE #363: Enable users to find the currently-selected object in the collection content tree
If users select an item, and then continue scrolling, it is easy to lose where that selected item was in the tree.

1. Implement darker focus/active styles for selected items.
2. Implement locate within collection button that scrolls back to the selected item on the page. Consider accessibility here. If the button only appears when the user scrolls away, is it necessary to announce to a SR that it has appeared? Maybe not, because a SR will not have the same disconnect between what is selected, and what is visually available on the screen.

Design solution: [https://projects.invisionapp.com/share/3Z10HL9DXPBK#/screens](https://projects.invisionapp.com/share/3Z10HL9DXPBK#/screens)

helrond:
Adds a button to scroll focused element back into view. Also changes color for focused items in tree.

I decided to have the button which scrolls the focused element into view always display, mostly because it's the less complex thing to do. If testing reveals that users can't see the button, then we can look at that as an enhancement.

Fixes #363

### DIFF
--- a/src/components/RecordsDetail/index.js
+++ b/src/components/RecordsDetail/index.js
@@ -143,6 +143,13 @@ const RecordsDetail = props => {
     props.params && props.params.query ? appendParams('/search', props.params) : '/'
   )
 
+  /** Scrolls a component into view in the records tree **/
+  const scrollFocusedIntoView = uri => {
+    const el = document.getElementById(`accordion__heading-${uri}`);
+    el.focus();
+    el.scrollIntoView({ behavior: 'smooth', block: 'center'});
+  }
+
   /** Parses an item's identifier from its URI */
   const identifier = (
     props.item.uri && props.item.uri.split('/')[props.item.uri.split('/').length - 1]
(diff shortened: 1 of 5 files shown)


HaSistrunk:
I'm cool with that. The only problem I see is that currently pressing the button from the top-level collection when a page loads (when nothing has focus yet) will throw an error: 
`TypeError: Cannot read property 'focus' of null`.

A thought: what about using the `btn--filter` styles for this instead? That one doesn't have any color change on hover, but typically our buttons darken on hover. I see the designs are using a new blue color (#DDE6F0), which I agree we don't need to add just for this button.

helrond:
OK, I added a conditional to catch cases where the element isn't in the DOM yet. I also did a bit of refactoring to the buttons, which was informed by thinking a bit more about making these styles reusable. I've done two things:
- Tried to consistently factor out color into its own class (`.btn--light-blue`)
- Use Sass `darken` function instead of applying a color. Over time, this approach would allow us to trim down the number of colors we have to articulate as hex codes.

HaSistrunk:
Nice, I like the `darken` approach.

Ok, I think we're at a good place to test this. I'm going to leave a note here to think through and research if/how AT users might use of the 'Locate within collection' button, and if it might need a different label or location in the DOM. I think it makes sense to see what we learn from usability testing first, though.

HaSistrunk (approved):

PR #369, opened 2021-03-31
Improves styling of mixed collection hierarchies

### LINKED ISSUE #362: Improve styles for collection content tree
When collections are mixed in with objects in the same level of the collection content tree, the styles are a little funky

We have a design solution for this: See [https://projects.invisionapp.com/share/4T10HL91ZEBV#/screens](https://projects.invisionapp.com/share/4T10HL91ZEBV#/screens).

helrond:
Improves styles for levels of collection hierarchies which include both objects and collections. Also adds logic to flip the Material Icon when accordions are expanded or collapsed.

Fixes #362

### DIFF
--- a/src/components/RecordsContent/index.js
+++ b/src/components/RecordsContent/index.js
@@ -23,6 +23,7 @@ export class RecordsChild extends Component {
     super(props)
     this.state = {
       children: [],
+      isExpanded: false,
       isSaved: false,
     }
   }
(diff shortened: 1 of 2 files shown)


HaSistrunk:
Nice. Thoughts on thickening the accordion div border when those have focus (the deep-navy)? I'm not sure it's necessary, but if we do it, it may be more easily accomplished using box-shadow: `box-shadow: 0 0 0 2px $deep-navy;`

image

helrond:
Hmm, that feels like it's slapping me in the face a little too much. I think I'd prefer to leave it as is?

HaSistrunk:
> Hmm, that feels like it's slapping me in the face a little too much. I think I'd prefer to leave it as is?

Yeah, that's fine. Looking at the designs, I think our borders are already a bit thicker anyway.

HaSistrunk (approved):

PR #381, opened 2021-06-08
Update URL even if API request fails

### LINKED ISSUE #380: Page Not Found URL should update even if there is an API request error
## Describe the bug
When the Page Not Found page opens because there is an error in the request to the API, the URL does not change.

## To reproduce
1. Navigate to a collection with an error (this bug was discovered from clicking on the last child collection under https://dimes.rockarch.org/collections/cvdH4EcvWPq9pXxRxDn7N8)
2. Click on collection or object
3. Notice when Page Not Found opens, the URL does not change

## Expected behavior
Even if the request returns an error, the URL should update to the target collection/object

## Impact on your work
This makes it difficult to report errors using specific URLs

helrond:
Moves `setURL` function to final `then` in axios call to make sure it's always called. 

Fixes #380

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "downshift": "^6.0.6",
     "formik": "^2.1.5",
     "mirador": "^3.0.0",
-    "node-sass": "^4.14.1",
     "pluralize": "^8.0.0",
     "react": "^16.13.1",
     "react-accessible-dropdown-menu-hook": "^2.1.1",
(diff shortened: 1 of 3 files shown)


HaSistrunk:
This looks ok to me, but I'm not sure how to actually test it now that the error I knew about has been fixed (and checks failing?).

helrond:
I'm not sure it's possible to test it without hardcoding a link that we know will 404. The build errors were related to `node-sass` (which is deprecated and replaced by `sass`).

HaSistrunk (approved):

PR #385, opened 2021-06-10
Improve logic for display of search results text

### LINKED ISSUE #383: No results heading displays before search results load
The "Sorry, there are no search results for..." heading displays before the actual search results load. If there is a slow connection, this can be confusing to the user.

If possible, adjust the timing so that this heading doesn't display unless there are actually no search results.

helrond:
Tweaks the logic for the search results text. While a search result is in progress, the text "Searching" will appear. Once the search has completed, either the text "Search results for {query}" or "Sorry, there are no results for {query}" will appear, depending on whether or not there were any results.

Fixes #383

### DIFF
--- a/src/components/PageSearch/index.js
+++ b/src/components/PageSearch/index.js
@@ -176,9 +176,12 @@ class PageSearch extends Component {
               category={this.state.params.category} />
           </div>
           <div className='results'>
-          <h1 className='results__title'>{this.state.resultsCount ?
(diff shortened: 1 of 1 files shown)


HaSistrunk:
Great. Is there an easy way to add the `loading-dots` class to just the inProgress state, or do you think that introduces too much logic that isn't worth it here?

HaSistrunk (approved):

PR #388, opened 2021-06-11
Add total results number to facet and a save & close button

### LINKED ISSUE #387: Provide indication that facets/filters are updating search results
Usability testing showed that users did not notice that facets/filters were updating the search results and facet/filter options when they selected the a checkbox input. We previously changed the design of the "Apply Dates" button because users were using this to apply filters, but this did not solve the confusion (see PR #368 ).

Add number of hits indication in modal that updates each time there are new results. Consider an aria-live property to announce the change via screen readers.

Previous discussions have included the idea of a loading indication for the search results page behind the modal.

helrond:
Adds the total results number to the Facets title. 

There are a couple of things that I think are maybe not ideal about this approach:
- Clicking on a facet doesn't always change the results count
- Depending on how far you've scrolled down the facet modal, the title text may not be visible
- There's a bit of a disconnect between the number of results (which are collections) and the number of hits (which are the actual matches.

Curious for your thoughts. It may be that we want to do a bit more research to figure out other UI patterns that work for this sort of thing. 

Fixes #387

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
-REACT_APP_ARGO_BASEURL=http://10.0.1.90:8010
+REACT_APP_ARGO_BASEURL=https://api.rockarch.org
 REACT_APP_REQUEST_BROKER_BASEURL=http://rac-vch.ad.rockarchive.org:8001/api
 REACT_APP_LOCALSTORAGE_KEY=DIMESMyList
 REACT_APP_S3_BASEURL=https://iiif.rockarch.org
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
-REACT_APP_ARGO_BASEURL=https://api.rockarch.org
+REACT_APP_ARGO_BASEURL=http://10.0.1.90:8010
 REACT_APP_REQUEST_BROKER_BASEURL=http://rac-vch.ad.rockarchive.org:8001/api
 REACT_APP_LOCALSTORAGE_KEY=DIMESMyList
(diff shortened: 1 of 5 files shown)


HaSistrunk:
The added bonus of this is that we could throw `aria-live='polite' aria-atomic='true'` as properties in the `h2` element, and screenreaders would state the number of results of they change when someone selects a checkbox. That would fix an existing usability problem for folks using assistive tech to traverse the modal.

That does leave the possible confusion around the disconnect between the number of results and the number of hits (which no one in testing has struggled with yet on the current site, but the information hasn't been presented to them in this more obvious way).

Idea: Keep the total results in the title (with the aria properties added), and also add a blue button at the bottom of the modal with the label "Save" that closes the modal. That may help with the need for folks to click on a thing to feel like they've safely applied something, but is not misleading like an "Apply" button.

helrond:
Cool cool, will make those accessibility changes - thanks for that! 

Building on the button idea, how about changing the close button at the top of the page with a blue button that says "Save" or "Save and Close"? That way we could avoid adding an additional UI element.

.env
```diff
@@ -1,4 +1,4 @@
-REACT_APP_ARGO_BASEURL=https://api.rockarch.org
+REACT_APP_ARGO_BASEURL=http://10.0.1.90:8010
```
HaSistrunk: We don't want to change this, right?

helrond: Good catch, thanks!

HaSistrunk (approved):

PR #389, opened 2021-06-11
Set category and query before performing search

### LINKED ISSUE #386: On search results page, digital filter checkbox reverts search term
As observed in usability testing, selecting the digital matches filter checkbox before submitting a search did not follow user expectations, resulting in the filtering of the results from their previous search.

## Steps to reproduce
1. Submit a search for any term
2. Now enter a new term, but do not submit the search. Instead, select the "show only results with digital matches" checkbox
3. Notice that the search term reverts to the previous term and applies the filter to that term.

I initially thought this was a bug, but in articulating the issue I realize that this is both about pattern consistency and understanding the action of the checkbox. Selecting it before submitting a search term is required from a homepage search, but it functions differently (as a post-search filter) on the search results page.

I have two ideas:

1. Do nothing. Error recovery was not a challenge as observed in testing, and the pattern is easy to learn even if it doesn't follow initial expectations. Problem was observed in one test, so track in future testing.
2. Change the checkbox functionality so that it is tied to the search submit button. I'm not sure how easy this is, but it also has the downside of removing the toggle functionality of the checkbox.

## Additional context
See PR #374.

helrond:
Sets the query and category correctly when the "Show only results with digital matches" checkbox is clicked. I also realized there was a bug in the logic for the `online` flag, so I've fixed that.

My only reservation here is that the select doesn't behave in the same way as the checkbox. Maybe I'm overthinking it, but that seems somewhat problematic?

Fixes #386

### DIFF
--- a/src/components/PageSearch/index.js
+++ b/src/components/PageSearch/index.js
@@ -128,8 +128,13 @@ class PageSearch extends Component {
   }
 
   /** Sets online flag when chechbox is checked **/
-  handleOnlineChange = (event) => {
-    var params = { ...this.state.params, online: event.target.checked }
+  handleOnlineChange = (event, query, category) => {
+    var params = { ...this.state.params, query: query, category: category }
+    if (event.target.checked) {
+      params = { ...params, online: true }
+    } else {
+      delete params.online
+    }
     this.executeSearch(params)
   }
 
--- a/src/components/PageSearch/index.js
+++ b/src/components/PageSearch/index.js
@@ -127,10 +127,11 @@ class PageSearch extends Component {
     this.executeSearch(params);
   }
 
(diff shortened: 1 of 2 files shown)


HaSistrunk:
> My only reservation here is that the select doesn't behave in the same way as the checkbox. Maybe I'm overthinking it, but that seems somewhat problematic?

Hm, I overlooked that initially but I agree that the lack of a consistent pattern is problematic. I tested this pattern quickly with three folks, and I think that the best approach is to make the select apply automatically/ initiate the search without having to additionally click on the search submit button so that it matches the function of the checkbox (and the facets).

helrond:
@HaSistrunk LMK how that looks!

HaSistrunk (approved):
Yep, looks great!

PR #391, opened 2021-06-28
Redo locate in collection and hover/focus styles changes

helrond:
When resolving merge errors, I mistakenly reverted the change which added the "Locate in Collection" button and improved the focus styles in container lists. I'm redoing this change here, but this should be reviewed to make sure I didn't bork something else in the process.


HaSistrunk:
I'm not seeing any problems. I did test on mobile a little more robustly, though, and the "Locate within collection" button has no real function when those sides of the screen are split. Do you think we should hide that button on small screens?

helrond:
That's a great call - let me do that right now.

HaSistrunk (approved):
Looks good to me!

PR #439, opened 2021-11-09
Updates pagination

helrond:
Updates to the next major version of `react-pagination`. There appear to be some minor differences in HTML structure:
- tabindex is -1 instead of 0 on pagination buttons
- some additional rel attributes on pagination buttons

Asking for a review in case any of these are dealbreakers for assistive technologies.

Release notes: https://github.com/AdeleD/react-paginate/releases/tag/v8.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "react-highlight-words": "^0.17.0",
     "react-loading-skeleton": "^3.0.1",
     "react-modal": "^3.11.2",
-    "react-paginate": "^7.1.3",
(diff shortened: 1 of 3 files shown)


HaSistrunk (commented):
This is all looking good to me and doesn't introduce any problems when I tested with VoiceOver. The tabindex change is an improvement for keyboard navigation because it doesn't focus on non-interactive things anymore.

HaSistrunk (approved):

PR #461, opened 2022-01-20
Uses ResizeObserver to trigger minimap render on page resize

### LINKED ISSUE #458: Handle resize of minimap container
## Describe the bug
The minimap uses a callback ref to calculate the size of squares. This div sometimes has an initial width or height of zero, causing the minimap to not render. I've added a `setTimeout` which seems to resolve this issue, but we should use a resize observer or a [library](https://www.npmjs.com/package/@react-hook/resize-observer) which wraps that API.

helrond:
Uses the ResizeObserver API to watch the Minimap container and re-render the minimap when the container size changes. Fixes #458

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "react-paginate": "^8.0.1",
     "react-router-dom": "^5.2.1",
     "react-scripts": "5.0.0",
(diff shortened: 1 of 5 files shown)


HaSistrunk (approved):
right on!

PR #462, opened 2022-01-20
Use correct justify-content value

### LINKED ISSUE #459: Minimap checkmarks are positioned differently in Safari
## Describe the bug
In Safari, the checkmark indicating that a minimap box has been clicked appears in the top left instead of the top right of the box as it does in other browsers.

## To reproduce
Open DIMES in Safari and search/open a collection with minimap results. Click on a box.

## Expected behavior
As designed and as seen in Chrome and Firefox, the checkmark appears in the top right of the square.

image

## Impact on your work
This is about design consistency, but has little to know impact on the function or use of the feature. Impact is very low. But since Safari is our second most commonly used browser to access DIMES, it makes sense that it be consistent.

helrond:
Fixes #459

### DIFF
--- a/src/styles/components/_minimap.scss
+++ b/src/styles/components/_minimap.scss
@@ -65,7 +65,7 @@ $grid-col-count: 20;
   background: $color;
   color: $color;
   display: flex; /* 1 */
-  justify-content: right; /* 1 */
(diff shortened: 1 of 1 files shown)


HaSistrunk (approved):
Good call. I'll correct this in the style library too.

PR #509, opened 2022-04-26
Improve display of sources, convert agent page subheaders to h3

### LINKED ISSUE #507: Clarify agent description sources
Usability testing showed that the sources of biographical information aren’t always clear on the agent pages. The bios come from the RAC, and when there is no RAC bio they are pulled from Wikipedia (if we have a Wikidata id). Currently at the end of the bio text we add `(Source: Wikipedia)` if it's from Wikipedia and do not add a source if it's from RAC. This approach was not clear to every usability test participant.

Include a source indication regardless of where the bio text is being pulled from. Put the source information ("Rockefeller Archive Center" or "Wikipedia") in its own paragraph in the "Description" section instead of a parenthetical at the end of the bio text:  

image

Secondarily, I'd advocate we make "Description" an `h3` instead of a paragraph. It's really functioning as a heading here. For example, see the `panel__heading` styles for the `h3`s in the Collection Details.

We could also move the source attribution up to the top of the bio text just below "Description." I'm not sure if this is necessary because it was clear from testing that people saw this information as a starting point that they would verify with other sources and likely not cite directly. However, this would clarify the source more prominently, particularly in the cases where the bio text is especially long.

cc @rpappous

helrond:
Wraps sources of agent notes in their own paragraph tag, and converts headings on agent attributes and notes to `h3`s. Some minor updates to styles and tests are also included.

Fixes #507

### DIFF
--- a/src/components/AgentAttribute/__tests__/AgentAttribute.test.js
+++ b/src/components/AgentAttribute/__tests__/AgentAttribute.test.js
@@ -2,10 +2,7 @@ import React from 'react'
 import { render } from 'react-dom'
 import AgentAttributeList from '..'
 
-const items = [
-  {'label': 'foo', 'value': 'foobar', 'note': true},
-  {'label': 'bar', 'value': 'barfoo', 'note': false}
-]
+const items = {'Positions Held': 'foo', 'Date of Birth': 'bar', 'Date of Dath': 'baz'}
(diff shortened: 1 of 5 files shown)


HaSistrunk (approved):
Looks good!

PR #510, opened 2022-04-26
correctly set initial isLoading value

### LINKED ISSUE #502: Add a loading indicator to My List page
When the My List page is loading, the user sees "No saved items." until the list loads. In cases where the list take some time to load, this causes confusion for the user.

Display the text "Loading..." until the list is loaded, similar to our "Searching..." and "Loading..." indicators for Search Results and Collection Contents .

Similar to the Search Results page "Searching..." loading indicator, we could make the "My List" h1 say "Loading...". It is probably preferable to have the page heading visible while it loads, though, so that users know what is loading.

## Additional context
A user in the Reading Room encountered this as an issue, thinking they were unable to request items when the page was taking a long time to load.

helrond:
fixes #502

### DIFF
--- a/src/components/PageMyList/index.js
+++ b/src/components/PageMyList/index.js
@@ -19,7 +19,7 @@
     const [savedList, setSavedList] = useState([])
     const [submitList, setSubmitList] = useState([])
     const [isDownloading, setIsDownloading] = useState(false)
-    const [isLoading, setIsLoading] = useState(false)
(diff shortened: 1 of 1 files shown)


helrond:
@HaSistrunk I know the tests are failing but that's fixed in #509, so I would propose merging this as-is.

HaSistrunk (approved):
That works!

Issue #469, opened 2022-02-01
Improve agent related collections


helrond:
## Is your feature request related to a problem? Please describe.
Agent related collections are currently returned from a basic keyword search against all collections. See if there are ways of getting better results back.

## Additional context
This might require additional API endpoints.

helrond:
Things to consider here:
- Whatever query is made here needs to be reproducible in the main search page
- Searching only for agents directly linked to collections will (at least now) overly limit the search results
- Doing a keyword search for a name might return non-relevant results (for example if a result has both names in a hit, but not necessarily together). At the same time, using a phrase search for the name will return fewer results since names are sometimes in inverted order in collection/object descriptions.

HaSistrunk:
Usability testing around this issue was inconclusive. For a current solution, what if we add the search term to the "Related Collections" heading to at least clarify how the matches are related?

Example:
For this URL: [https://dimes.rockarch.org/agents/FKCXNeQe67sggaQcME8fFe?category=&limit=40&query=music%20education](https://dimes.rockarch.org/agents/FKCXNeQe67sggaQcME8fFe?category=&limit=40&query=music%20education) the header would be:

<img width="628" alt="image" src="https://user-images.githubusercontent.com/15254041/165334157-26dc99d2-6b71-4ec5-851c-e2b4fdf9e129.png">

cc @rpappous

PR #512, opened 2022-04-26
Use agent name as query

helrond:
Uses the agent name as a query for related collections, as opposed to whatever query string brought a user to the agent page. Also adds the agent name to he header for related collections so it's clear what the collections are related to.

Not a full solution, but towards one for #469


HaSistrunk:
In testing this, I'm not seeing the agent title in the header.

HaSistrunk (approved):
Ok cool, this is working (I deleted a previous comment that I wasn't seeing the agent title in the heading - I just needed to pull recent changes)

PR #537, opened 2022-07-29
Throw more obvious errors when calls to backend APIs fail

### LINKED ISSUE #236: Home page text is not centered on larger screens

### LINKED ISSUE #536: Report error in UI when API calls fail
When calls from DIMES to api.rockarch.org fail or time out, users just see endless loading in DIMES. We've seen from several researchers in the Reading Room (and staff attempting to help them troubleshoot) that it isn't clear to users what the problem is or why the page won't load, leading them to think it may be their own browser's or computer's error.

When API calls fail, display an error message alerting the user that there is a problem with the system.

The error message when the Request Broker is unavailable is: "Sorry, our system is unable to process requests right now. We're working to fix this! Please try again later." I'd recommend a similar message appear on the page for failed API calls:
"Sorry, our system is unable to retrieve the information you requested right now. We're working to fix this! Please try again later." 

This would appear on every page where a failed API call would prevent the page from loading:
1. Agent
2. MyList
3. Records
4. Search

Consideration: I'm wondering if stating "We're working to fix this!" would discourage our staff users from reporting the problem?

helrond:
Show an error page when calls to backend APIs (Argo/Request Broker) fail. 

Fixes #536

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,40 @@ script:
   - docker build .
   - CI=false yarn run build
   - yarn test
+before_deploy:
+  - sudo deploy_scripts/make_zip_react.sh $DEPLOY_ZIP_DIR $DEPLOY_ZIP_NAME
 deploy:
+# - provider: s3
+#   access_key_id: $AWS_ACCESS_KEY
+#   secret_access_key: $AWS_SECRET_KEY
+#   local_dir: $LOCAL_DIR
+#   skip_cleanup: true
+#   on:
+#     repo: RockefellerArchiveCenter/${APPLICATION_NAME}
+#     branch: base
+#   bucket: $AWS_BUCKET_NAME
+#   region: us-east-1
 - provider: s3
   access_key_id: $AWS_ACCESS_KEY
   secret_access_key: $AWS_SECRET_KEY
-  local_dir: $LOCAL_DIR
+  local_dir: $DEPLOY_ZIP_DIR
   skip_cleanup: true
   on:
     repo: RockefellerArchiveCenter/${APPLICATION_NAME}
     branch: base
   bucket: $AWS_BUCKET_NAME
   region: us-east-1
+- provider: codedeploy
+  bucket: $AWS_BUCKET_NAME
+  key: $DEPLOY_ZIP_NAME
+  bundle_type: zip
+  access_key_id: $AWS_ACCESS_KEY
+  secret_access_key: $AWS_SECRET_KEY
+  application: ${APPLICATION_NAME}
+  deployment_group: DimesProduction
+  region: us-east-1
+  on:
+    repo: RockefellerArchiveCenter/${APPLICATION_NAME}
+    branch: base
 notifications:
   email: false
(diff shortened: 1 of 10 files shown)


HaSistrunk:
The error page/message looks good to me. A few things/questions:

- One thing I'm seeing (but I can't figure out why in the code) that seems to have been impacted by this change is that the inProgress loading indicators when you search DIMES are no longer appearing. Instead, while it loads search results it shows the page indicating that there are no results.

1. API calls are working with no errors
2. Search for anything in DIMES
3. See the "Sorry, there are no search results" page
4. See that there are actually search results that then load.

- For the records page, the 404 page is the one that displays when the API call fails, which is the same as it is in the `base`. Am I missing the intended change there? I'm testing locally with the API URL in the `.env` pointing to a wrong URL.

- There are no intended UI changes with Request Broker errors in this PR, right? I see the same modal error message that already exists.

helrond:
OK, thanks for this review!
- I fixed the first issue - just needed to wait for the response and then set `inProgress`
- The API call can fail in multiple ways. When the response is a 404, you should get the 404 page. When it's another kind of error (you get a 500 response or no response at all) you'll get the PageBackendError (at least that's how it's supposed to work). If that's NOT what's happening can you give me some more details on URLs?
- No changes have been made to the text in the MyList modal. That text is determined by a call to the `status` endpoint of the Request Broker, so it's effectively testing whether the app is available. As @ctgraham's comment in #536 indicates, there are other Request Broker calls that can fail, which are the result of unexpected data or other application bugs. In those cases we're just going to show the PageBackendError because those issues are dealbreakers and need to be resolved before anything can happen with the data in the MyList page.

HaSistrunk:
> * The API call can fail in multiple ways. When the response is a 404, you should get the 404 page. When it's another kind of error (you get a 500 response or no response at all) you'll get the PageBackendError (at least that's how it's supposed to work). If that's NOT what's happening can you give me some more details on URLs?

It looks like now both the 404s and the no response errors both show the 404 page. For example: `/objects/F2p2iUQgey2hiRX7ZMEWcv` or `/collections/KVPbAKnCc5MF7STt99VEGf` exist, but when I try to load those pages without the API (call fails) it loads the 404 page. I testing that with a few different records and see the same.

HaSistrunk:
> > * The API call can fail in multiple ways. When the response is a 404, you should get the 404 page. When it's another kind of error (you get a 500 response or no response at all) you'll get the PageBackendError (at least that's how it's supposed to work). If that's NOT what's happening can you give me some more details on URLs?
> 
> It looks like now both the 404s and the no response errors both show the 404 page. For example: `/objects/F2p2iUQgey2hiRX7ZMEWcv` or `/collections/KVPbAKnCc5MF7STt99VEGf` exist, but when I try to load those pages without the API (call fails) it loads the 404 page. I testing that with a few different records and see the same.

Ok, I can confirm that this is actually working. The issue was in how I was testing this locally, not the code. So this looks good to go to me!

HaSistrunk (approved):

PR #542, opened 2022-08-05
Update argo docs link

helrond:


HaSistrunk:
@helrond it looks like the URL for this is still `docs.rockarch.org/argo/`, not `argo-docs`

helrond:
Internally that's true, but externally it is `argo-docs`. This is a temporary thing while we get some networking things sorted out.

Issue #563, opened 2022-11-10
Page titles not consistently sent to Matomo Analytics


HaSistrunk:
## Describe the bug
We recently switched from using Google Tag Manager to Matomo Tag Manager (PR #552). In implementing, the `firePageViewEvent` that we created to address Issue #339 (added in PR #349) is not working with Matomo Tag Manager, causing the same issue that we had before with page titles not consistently being set before page title data is available to Tag Manager, which results in page titles that default to the "DIMES: Online Collections and Catalog of Rockefeller Archive Center" that is in index.html.

## Testing
In Matomo Analytics Behavior section, compare "Page Title" report with "Pages" report on 11/10/22 or later (date of Matomo implementation), and see that the `/index` URL has far fewer pageviews than the "DIMES: Online Collections..." page title pageviews.

Use the debug/preview feature in Matomo Tag Manager in the DIMES container to test tags and view the logs and datalayer to see event data.

## Impact on your work
This impacts the accuracy of analytics reports that include page titles, although the page URLs are accurate.

## Additional context
There are a few options for what is going on:
1. The `firePageViewEvent` is not correctly implemented in the DIMES code to push to the [Matomo datalayer](https://developer.matomo.org/guides/tagmanager/datalayer).
2. The `firePageViewEvent` is not set up in Matomo Tag Manager correctly to trigger the pageview tag
3. There is possibly another solution for Matomo Tag Manager that is different than the Google Tag Manager solution. See [Matomo Docs on tracking SPAs
](https://developer.matomo.org/guides/spa-tracking).

HaSistrunk:
I'm still troubleshooting this. I have:

1. Confirmed in Matomo debug/preview mode in the Logs tab that the event `reactPageViewEvent` is being pushed to the Matomo datalayer, so the event seems to be correctly implemented in DIMES based on the "Triggering an Event" [Matomo instructions](https://developer.matomo.org/guides/tagmanager/datalayer).
2. Following the "[Configuration in Matomo Tag Manager](https://developer.matomo.org/guides/tagmanager/datalayer#configuration-in-matomo-tag-manager)" instructions, I 
created a trigger called "Pageview - Trigger React Event" of the type `Custom Event`, and assigned `Event Name` the value `reactPageViewEvent`. I then created a pageview tag called "Pageview - React Event" that uses this custom pageview trigger.

This pageview tag _should_ fire when a new page is loaded, but from the debug/preview mode, I see that it is not.

PR #564, opened 2022-11-15
Debug reactPageViewEvent for Matomo

HaSistrunk:
Fixes issue #563 

Debugs the custom pageview event helper to work with Matomo: window.dataLayer referenced Google's datalayer when it should have been pointing to Matomo.

I tested this in Matomo's Tag Manager preview and it appears to be working and sending the correct page titles as expected.

@helrond , what is line 85 accomplishing in `Helpers.js`? Trying to make sure I understand all the elements here.

### DIFF
--- a/src/components/Helpers.js
+++ b/src/components/Helpers.js
@@ -82,11 +82,11 @@ export const firePageViewEvent = title => {
     done = false
   }
   if (title && !done) {
-    if (window && window.dataLayer) {
-      let dataLayer = window._mtm || []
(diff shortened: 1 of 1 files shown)


helrond:
Line 85 is checking for the existence of the `window._mtm` variable (technically it's checking the existence of both `window` and `window._mtm` so that we don't try and send the page event if that variable is missing, for example in testing environments.

helrond (approved):

PR #600, opened 2023-05-08
Issue 545/lingui translation wrappings

### LINKED ISSUE #545: Localization and Internationalization of the UI
As others adopt this application, they will need a way to easily edit the text in the UI to reference their own institution's name, contact, policies, etc.

I propose using i18n / l10n modules to address both the use cases of translation and customization.  Abstracting the text to translation files allows the implementor a single area of concern for editing the UI.

- https://phrase.com/blog/posts/react-i18n-best-libraries/
- https://aglowiditsolutions.com/blog/react-internationalization-localization/

Other solutions haven't really been considered; we don't have any React experience in house.  This Issue is intended to generate the discussion to scope and strategize.

## Additional context
This leaves outstanding the question of branding, which I see as a distinct challenge for implementors.

Tyl13:
This is my current work that is done on adding Lingui Translation Wrappings. It includes changes to yarn.lock. New messages are extracted by running yarn extract and then they are compiled to Javascript with yarn compile. I have a few questions about certain sections but if you have anything that you want clarified first, or just more information on what is happening. Please ask. When this is finished, it will resolve #545. I believe there are some questions in the issue too that may  provide additional information about Lingui.


src/components/i18n/index.js
Tyl13, line 1: This file actually should have been removed, it is how I was doing dynamic activation of the different languages while testing. This can stay, and the only use of it being to activate the English locale.

Tyl13: I additionally have the dynamic activation code on a different branch if you want that included too.

lingui.config.js
Tyl13, line 1: This is the current settings for lingui. They can also be placed within a file name .linguirc, a lingui section within package.json or lingui.config.ts. If you have a preference for where they should go instead, I can change them.

lingui.config.js
```diff
@@ -0,0 +1,10 @@
+/** @type {import('@lingui/conf').LinguiConfig} */
+module.exports = {
+	locales: ["en", "fr"],
```
Tyl13: The fr locale is just being used as a dummy locale for testing. It can be removed without issue and lingui will still function without it. I can also change the language to one that you know you want to start translating to.

lingui.config.js
```diff
@@ -0,0 +1,10 @@
...
+	sourceLocale: "en",
```
Tyl13, line 4: This is the language that the messages are originally in. This language also will not require any translation work unless you want something changed since it pulls the message straight out into locales/en/messages.po.

lingui.config.js
```diff
@@ -0,0 +1,10 @@
...
+	catalogs: [{
+	   path: "src/locales/{locale}/messages",
```
Tyl13: This is where messages are located, this can also be changed to have messages to be next to each component, not within its own locales and just named the locale.po, and many more choices. I do believe that most translators would rather have one big file to translate since it is easier to not miss anything. This however causes the file to grow larger and larger.

I do see some benefits to separating them either into `{locale}/{component}/messages` or `src/components/{component}/locales/{locale}/messages` but I'm not sure if they are worth it.

On one side having a giant messages allows all translations to stay in one spot, not worry about having to track each translation file down, multiple files can rely on only one string to be changed to change the displayed string in multiple places.

On the other side has smaller messages, no issues where two matching strings should have two different translations because of context, and keeps component information together.

lingui.config.js
```diff
@@ -0,0 +1,10 @@
...
+	   include: ["src"]
+	}],
+	format: "po"
```
Tyl13, line 10: The format is by default po, which is used by the GNU gettext. They also have json and csv format options included, but custom formatters can also be used.

src/components/Facet/index.js
```diff
@@ -3,15 +3,31 @@ import PropTypes from 'prop-types'
...
+        t({
+          comment: 'Message shown when list is opened',
+          message: 'show less'
+        })
+        :
+        t({
+          comment: 'Message shown on list is closed',
+          message: 'show all'
+        })
+      }
```
Tyl13: What are you thoughts on doing conditional logic like this for which translation to show. The other option is to remove the conditional outside of the translation and use a select within the message for conditional logic. I have an example of that [here](https://github.com/RockefellerArchiveCenter/dimes/blob/16303b04d302bef83dacebe34b7635aa71785b02/src/components/MyListExportActions/index.js#L19-L24) which I also commented below.

src/components/Footer/index.js
```diff
@@ -1,35 +1,123 @@
...
       <div className='wrapper'>
         <div className='container'>
-          <h2 className='footer-primary__title'>Rockefeller <span className='line-break'>Archive Center</span></h2>
+          <h2 className='footer-primary__title'>
+            <Trans comment='Primary Title'>
+              Rockefeller
+              <span className='line-break'>
+                Archive Center
+              </span>
+            </Trans>
```
Tyl13, line 19: Do you want each individual connected item to be translated, or do you want the whole file wrapped by a Trans to convert whole components at a time?

Tyl13: I think for the most part individual unconnected items is better, but some sections like the Social Icons could probably be one huge Wrappings since there are so few actual strings that need translated and they all seem related.

src/components/Footer/index.js
```diff
@@ -1,35 +1,123 @@
...
+              <a className='footer-primary__link footer-primary__policy-link' href={
+                t({
+                  message: 'https://docs.rockarch.org',
+                  comment: 'Footer.Primary.RAC.href'
+                })
+              }>
+                <Trans comment='Footer.Primary.RAC.message'>
+                  RAC Policies
+                </Trans>
+              </a>
```
Tyl13: Do the translation wrappings create too much clutter compared to the original? Does it need cut down, and do you have a preference on how it is reduced?

src/components/Helpers.js
```diff
@@ -51,9 +52,17 @@ export const formatBytes = (bytes, decimals = 2) => {
...
-  const suffix = online ? (hitCount === 1 ? 'digital match' : 'digital matches' ) : (hitCount === 1 ? 'match' : 'matches')
+  const physicalMatch = t({
+    comment: 'Pluralization of match(es)',
+    message: plural(hitCount, {one: 'match', other: 'matches'})
+  })
+  const digitalMatch = t({
+    comment: 'Pluralization of digital match(es)',
+    message: plural(hitCount, {one: 'digital match', other: 'digital matches'})
+  })
+  const suffix = (online ? digitalMatch : physicalMatch)
```
Tyl13, line 65: This is doing different conditional logic but when it pertains to the number of items. Doing it this way also allows for the pluralization to be extended for languages that have more than two types of plural forms.

src/components/ListToggleButton/__tests__/ListToggleButton.test.js
```diff
@@ -26,7 +27,7 @@ it('renders correctly in desktop and modal', () => {
   })
 
   const button = document.querySelector('button')
-  expect(button.textContent).toContain('Add to List')
+  expect(button.textContent).toContain(t({comment: "Test function", message: 'Add to List'}))
```
Tyl13: If I can get the actual lingui working using a wrapper, should I translate the tests too? Or should I not try to get that working and just mock lingui?

src/components/ListToggleButton/__tests__/ListToggleButton.test.js
Tyl13, line 1: I'm going to use this file to represent testing as a whole. As reported by travis, the tests fail. That is currently because no I18nprovider is being provided to the render. I can fix this by mocking multiple parts of the lingui. I'm still currently working on fixing one more issue with the mocking. However, since I'm mocking it. I am just returning the original string. So the two original strings need to stay in sync or else it'll fail. 

I was hoping to be able to translate both the expected and original string at once. This would allow the expected string and original string to stay the same without extra work. It would also allow to make sure the actual translation was working. 

I think there is a way to use a wrapper on the render being used for testing but I haven't figured that out yet.

src/components/Minimap/index.js
```diff
@@ -45,14 +46,22 @@ const Minimap = ({ data, isLoading, params, rowCount=4 }) => {
...
     const areaUrl = areaHits.length && areaHits[0].uri
     const isAreaActive = areaUrl === currentUrl
+    const hitPlural = t({
+      comment: "Pluralization of hit(s) for minimap boxes",
+      message: plural(areaHits.length, {one: "hit", other: "hits"})
+    })
+    const message = t({
+      comment: "Message displayed to jump to specific section within minimap",
+      message: `Jump to ${areaHits.length} ${hitPlural} in this area: ${hitTitles}`
+    })
```
Tyl13, line 56: I made the messages that I needed translated done outside of the returned data. Is this a better way for me to make the code easier to read?

src/components/Minimap/index.js
```diff
@@ -45,14 +46,22 @@ const Minimap = ({ data, isLoading, params, rowCount=4 }) => {
...
-        key={idx}
-        href={appendParams(areaUrl, params)}
-        className={classnames('minimap__box', hitClass, { [`${hitClass}--active`] : isAreaActive})}
-        title={`Jump to ${areaHits.length} ${areaHits.length === 1 ? 'hit' : 'hits'} in this area: ${hitTitles}`} >
-        <span className='visually-hidden'>{`Jump to ${areaHits.length} ${areaHits.length === 1 ? 'hit' : 'hits'} in this area: ${hitTitles}`}</span>
+          key={idx}
+          href={appendParams(areaUrl, params)}
+          className={classnames('minimap__box', hitClass, { [`${hitClass}--active`]: isAreaActive })}
+          title={message} >
+        <span className='visually-hidden'>{message}</span>
```
Tyl13, line 64: This is what it looks like within the code, when I define the translation outside of the return and use it within it.

src/components/ModalMinimap/index.js
```diff
@@ -3,20 +3,21 @@ import PropTypes from 'prop-types'
...
+import { t, Trans, Select } from '@lingui/macro'
 import './styles.scss'
 
-const minimapAboutText = <>
+const minimapAboutText = <Trans comment='The about section for minimaps.'>
   <p>Jump to a part of the collection containing matches by clicking on an active square.</p>
   <p>In the minimap diagram, each square represents part of this collection.
   Colored squares represent parts that contain one or more matches for your search.</p>
-</>
+</Trans>
```
Tyl13, line 13: Do you have a preference for me replacing empty divs with the Trans macro? In these spots, it seemed to make sense to not have an empty div around the Trans macro.

src/components/ModalSearch/index.js
```diff
@@ -40,11 +41,16 @@ export const FacetModal = props => {
       closeTimeoutMS={200} >
       <div className='modal-header--search'>
         <h2 className='modal-header__title' aria-live='polite' aria-atomic='true'>
-          {`Filter ${props.resultsCount} Search ${props.resultsCount === 1 ? 'Result': 'Results'}`}
+          <Trans comment='Filter results message'>
+            <Plural value={props.resultsCount} one="Filter # Search Result" other="Filter # Results"/>
+          </Trans>
```
Tyl13, line 46: The Trans macro is around the Plural macro to easily assign a comment to the plural. Additionally, the `#` will be replaced with the value after the translation.

src/components/PageBackendError/index.js
```diff
@@ -25,7 +28,7 @@ const PageBackendError = ({error}) => (
           <p className='backend-error__message'>{error.config.data}</p>
         </> : null}
     </div>
-  </>
+  </Trans>
```
Tyl13, line 31: This is a section where I am not sure if replacing the empty div with the Trans macro is better or not. This is also a section where I'm not sure if the whole translation should be one big item or multiple smaller ones.

lingui.config.js
helrond: I agree with your instinct here - let's keep each locale in one file. We can always pivot later if it becomes necessary.

lingui.config.js
helrond: I think I am inclined to remove the fr locale before we merge. It's useful to see how it all hangs together but I think I'll need a bit of time to figure out what translations we want to put into place and who will do that work.

lingui.config.js
helrond: I don't have a strong preference here, so I think we can leave this where it is.

src/components/Facet/index.js
helrond: Hmmm...at first glance the `select` method seems a lot cleaner. Do you have a preference/thoughts from working through this?

src/components/Footer/index.js
helrond: I think that's right - if someone were going to add or remove one of these social icons they'd have to make changes to the HTML structure anyway.

src/components/Footer/index.js
```diff
@@ -1,35 +1,123 @@
...
+              <Trans comment='Full Company Address'>
+                15 Dayton Avenue
+                <br />
+                Sleepy Hollow, New York
+                <br />
+                10591
+              </Trans>
+            </p>
+            <p className='footer-primary__text'>
+              <Trans comment='Comapny Phonenumber'>
```
helrond:
```suggestion
              <Trans comment='Company Phonenumber'>
```

src/components/Facet/index.js
```diff
@@ -3,15 +3,31 @@ import PropTypes from 'prop-types'
...
+      label={
+        isOpen
+        ? 
+        t({
+          comment: 'Message shown when list is opened',
+          message: 'show less'
+        })
+        :
+        t({
+          comment: 'Message shown on list is closed',
```
helrond:
```suggestion
          comment: 'Message shown when list is closed',
```

src/components/Footer/index.js
helrond: It is a lot more cluttered for sure. I guess one approach would be to create a PrimaryLink component?

src/components/Footer/index.js
```diff
@@ -40,16 +128,52 @@ const Footer = () => (
...
+                <a className='footer-secondary__link' href={
+                  t({
+                    comment: 'Take-down Policy link',
+                    message: 'https://docs.rockarch.org/takedown-policy/'
+                  })
+                }>
+                  <Trans comment='Take-down Policy message'>
+                    Take-down policy
+                  </Trans>
+                </a>
```
helrond: If we're going to go the route of creating a PrimaryLink component, I think we could potentially do the same here for SecondaryLink.

src/components/ListToggleButton/__tests__/ListToggleButton.test.js
helrond: I would be very happy if the tests also included translation, so if you can get that working I am all for it!

src/components/Minimap/index.js
helrond: Yes, I like this!

src/components/Minimap/index.js
helrond: Yeah I think it's not always necessary, but in this case it really does clean things up a lot.

src/components/ModalMinimap/index.js
helrond: Agree - the only reason those empty divs are there is to satisfy React.

src/components/MyListExportActions/index.js
```diff
@@ -2,22 +2,35 @@ import React from 'react'
...
     <Button
       className={classnames('btn--orange', 'btn--sm', {'loading-dots': isDownloading})}
-      label={isDownloading ? 'Downloading' : 'Download as .CSV'}
+      label={t({
+        comment: 'Downloading label',
+        message: select(isDownloading, {
+          true: 'Downloading',
+          other: 'Download as .CSV'
+        })
+      })}
```
helrond, line 25: As noted above, this feels a lot cleaner, and keeps all the translation stuff together neatly.

src/components/PageBackendError/index.js
helrond: I agree with what you've done here. If someone sees this page we really want them to screengrab it and send it to us. I don't think the empty divs serve any purpose other than keeping React happy because of the Helmet component.

src/components/Footer/index.js
HaSistrunk: I like the idea of decluttering with PrimaryLink and SecondaryLink components as part of this PR.

HaSistrunk:
This is looking really solid! I'm wondering if we need a README update to document/point to documentation on how to implement translations.

Tyl13:
Hopefully nothing was missed and it all looks good, but it should be good to merge.

helrond (approved):
lgtm 🚀

HaSistrunk (approved):

PR #601, opened 2023-05-17
Implement BackstopJS for Visual Regression Testing

HaSistrunk:
Adds [backstopJS](https://github.com/garris/BackstopJS) for visual regression testing including scenarios, reference images, and mock data in the `/request_reference` directory. This implementation uses the Backstop Docker container to run tests consistently across different environments. 

It also updates the README to include instructions for running tests and managing reference images.


README.md
```diff
@@ -15,6 +15,17 @@ Install dependencies and run the development server:
...
+### Visual regression testing
+
+The repository includes [BackstopJS](https://github.com/garris/BackstopJS) to test visual changes to the site by comparing a set of reference images for different screen sizes. Anytime the CSS styles are changed, use BackstopJS to test locally:
+
+1. Run the BackstopJS tests: `yarn backstop-test`.
+2. Review the results in the browser and look at the diff of any failed tests.
+3. To update the reference image files with the results of the last test images use: `yarn approve`. Subsequent tests will be compared against these updated reference files.
+4. Commit any updated reference images to the repository so that future tests will be compared against the most recent images.
+
+To add or update reference images, edit the scenarios in `backstop.json` and run `yarn reference`.
```
helrond: I wonder if we want to use a similar format for all backstop commands, like `backstop-test`, `backstop-approve` etc rather than only using the `backstop-` prefix in one place. I know it's more to type but it makes these commands somewhat easier to guess.

README.md
```diff
@@ -15,6 +15,17 @@ Install dependencies and run the development server:
     $ yarn install
     $ yarn start
 
+### Visual regression testing
+
+The repository includes [BackstopJS](https://github.com/garris/BackstopJS) to test visual changes to the site by comparing a set of reference images for different screen sizes. Anytime the CSS styles are changed, use BackstopJS to test locally:
+
+1. Run the BackstopJS tests: `yarn backstop-test`.
```
helrond: Do we need to note that the development server needs to be running in another terminal using `yarn start`?

helrond (commented):
I ran this locally and everything worked EXCEPT the "Search Sort Dropdown" scenario, which timed out while waiting for the `#downshift-1-toggle-button` selector. I also see that you don't have any reference images for that scenario though, so were you planning on removing it?

I've also added some specific comments around documentation.

HaSistrunk:
> I ran this locally and everything worked EXCEPT the "Search Sort Dropdown" scenario, which timed out while waiting for the `#downshift-1-toggle-button` selector. I also see that you don't have any reference images for that scenario though, so were you planning on removing it?
> 
I'm not sure why, but the selector for that one changed to `#downshift-4-toggle-button`, so I updated it. The reference files for that one are hiding in there at the end.

helrond (approved):
Thanks for those tweaks - looks good to me and the BackstopJS tests pass now!

PR #603, opened 2023-05-24
Implement Style Library

HaSistrunk:
Implements the RAC Style Library and Guide into DIMES: 
- Uses RAC Style Library stylesheet from CDN, replacing some local styles
- Updates color variable names and removes some colors to match Style Library
- Updates HTML for some elements to match RAC [Style Guide](styles.rockarch.org).
- Removes Neat and uses CSS grid for some layouts
- Where possible, uses spacing utility classes for margin and padding in the HTML instead of defining in the stylesheets.
- Updates BackstopJS scenario selectors to work with new HTML structures.

There are no major visual changes in the UI, but there are some minor changes that are apparent in the visual regression tests. From my perspective the changes are so slight that they don't matter, are improvements from the original, or are acceptable because they make the code more maintainable. I'm open to disagreements on any of these calls!

I've left the BackstopJS Reference files to be used for review, but I'll update them once we're happy with this PR before merging.


HaSistrunk:
A few notes:
- Right now, the sass styles that have been removed/replaced are just commented out. We'll want to delete those before merging.
- I haven't run BackstopJS yet for visual regression testing. We'll need to do that, and also review the scenarios and update the Reference images before merging because there will be some small visual changes.

There are also a couple things that aren't looking like they should:
1. This isn't lining up as it should, but I can't figure out what changed:
image

2. On mobile, the header nav dropdown doesn't extend all the way down the page. Gotta fix that.
image

HaSistrunk:
> 1. This isn't lining up as it should, but I can't figure out what changed:
> 
> image

This was due to the `box-sizing` value, and should now be fixed with a change to the `child__list` class to use `content-box` instead of the inherited `border-box`.

> 2. On mobile, the header nav dropdown doesn't extend all the way down the page. Gotta fix that.

Fixed by removing the local dimes position styles from the `dropdown__list` class and applying them via another class (`mylist__actions--dropdown`) only where they are needed on the mylist page dropdown.

src/App.js
```diff
@@ -55,7 +55,6 @@ const App = () => {
   return (<LiveAnnouncer>
     <SkipLink />
     <Header myListCount={myListCount} />
-    <main id='main' role='main'>
```
helrond, line 58: This appears to break the skip link functionality, which is pointing to `#main`...

helrond: Oh, ugh I see some page components have `#main` and others don't...

HaSistrunk: Oh I missed that. I thought there were two `main` landmarks on all pages so I deleted that, but I see we need an approach to make sure there is one on every page...

helrond:
Another thing that's occurring to me is that this is actually a pretty impactful change in terms of HTML structure and _may_ cause a lot of merge issues for other PRs which broadly impact application code, like the translation PR. I'm not sure what the best way is to handle that; someone's going to have to deal with the merge mess.

HaSistrunk:
> Another thing that's occurring to me is that this is actually a pretty impactful change in terms of HTML structure and _may_ cause a lot of merge issues for other PRs which broadly impact application code, like the translation PR. I'm not sure what the best way is to handle that; someone's going to have to deal with the merge mess.

Since this isn't really a styles implementation change, I'm wondering if it makes sense to hold off on this change and file the double `main`s as an issue to address separately?

helrond:
I was thinking more about this overall PR (less about the `main` issue) - let me think some more about this but I think we might need to do a bit of coordination here.

HaSistrunk:
Two questions:
1. Lato italic font has different spacing now with the imported stylesheet (most visibly in the header), but I can't figure out why it changed. Any ideas?
2. How do you feel about the use of the margin and padding utility classes? It was implemented in this [styles PR](https://github.com/RockefellerArchiveCenter/styles/pull/139), and if we decide to go with it I'll make sure it's documented in Storybook there as well.  In general, I like this approach and I think it will make implementation easier on some of our other sites that use Bootstrap. The downsides I see are:
    - That they can't handle media queries,  so spacing that changes with screensize has to be handled in the SCSS. 
    - It adds a bunch of lines to the stylesheet to create each class from 0 to 60 px for margins and padding.

helrond:
My guess on the Lato font difference is that the imports are different - in `styles` you're [importing Lato Italic](https://github.com/RockefellerArchiveCenter/styles/blob/base/stylesheets/base/_typography.scss#LL14C5-L14C5) but in DIMES [we're just importing Lato](https://github.com/RockefellerArchiveCenter/dimes/blob/base/src/styles/base/_typography.scss#LL1C1-L2C1). That said, I personally think that the version in `styles` looks better...

If the spacing classes are useful in this repo, my guess is that they'll be quite useful globally, particularly when it comes to updating Bootstrap sites (as you note). One thought - rather than make these based on `px` could we use `rem` as a unit? That would potentially make them more responsive.

helrond (approved):

HaSistrunk:
> One thought - rather than make these based on `px` could we use `rem` as a unit? That would potentially make them more responsive.

I looked more comprehensively at where we use media queries for spacing in DIMES and where there are other margin/padding styles that can't be accomplished with utility classes:
- Media queries are used primarily for big responsive changes like the hero on the homepage where the margin changes from 10px to 113px (where rems wouldn't be useful). We also use them for responsiveness in typography in the styles library, but using rems there feels like a separate project that would accompany a shift in font-size units as well.
- We also have margin/padding styles defined in the SCSS when the selectors are more complex and styles can't be applied in the HTML easily, like `:first-of-type` or `input[type=number]`.

So I'd vote to leave things as they are in this PR, but keep a spacing and typography refactoring project in mind for possible future work.

src/styles/components/_collection-hits.scss
HaSistrunk, line 1: @helrond , we aren't using these `collection-hits.scss` styles in DIMES currently. I think they are from a previous design iteration? Do you think it's still important to hang onto them, or can we delete?

helrond: Your memory matches mine - I think this is an abandoned design idea. Let's get rid of these.

helrond (approved):
Looks good!

PR #608, opened 2023-06-07
Reading room date picker

chryslovelace:
This uses the endpoint added to request-broker in https://github.com/RockefellerArchiveCenter/request_broker/pull/299 to get name and time information about available reading rooms from Aeon in order to ensure the user selects a valid time.


chryslovelace:
As it stands, the code here doesn't allow for a way to bypass the Aeon API integration that pulls in the reading room data. Is that something you would want, and if so, how would you want that to be configured?

helrond:
@chryslovelace Right now we don't want users to have to pick a Reading Room or a time in our implementation of DIMES, so yes, it's something I'd like to be able to turn off. So far we've put all configuration stuff in `.env` but I'm wondering if there are better solutions for this that you'd suggest?

helrond:
Hi @chryslovelace - I'm gonna trust that the implementation for when you have reading rooms set up in Aeon works because we don't have that set up right now. HOWEVER I am seeing that the `REACT_APP_ENABLE_READING_ROOM_SELECT` is not working as expected, which I think is because that variable is being parsed as a string, not a boolean, so as long as it's present it evaluates to `true`. There are probably a number of ways to address this (only add the environment variable if you want to enable the reading room select, convert the string to a boolean, etc) but regardless of which you go with we should probably document this feature in the README.

chryslovelace:
As suggested, I converted that env var to boolean where its used, so it'll be activated as long as the env var is present and is anything but the empty string. The build process seems to be getting stuck in some sort of loop.

.env
```diff
@@ -3,4 +3,5 @@ REACT_APP_REQUEST_BROKER_BASEURL=http://rac-vch.ad.rockarchive.org:8001/api
 REACT_APP_LOCALSTORAGE_KEY=DIMESMyList
 REACT_APP_MINIMAP_KEY=DIMESMinimapIntro
 REACT_APP_S3_BASEURL=https://iiif.rockarch.org
-REACT_APP_CAPTCHA_SITE_KEY=6LdQiSkTAAA
\ No newline at end of file
+REACT_APP_CAPTCHA_SITE_KEY=6LdQiSkTAAA
+REACT_APP_ENABLE_READING_ROOM_SELECT=false
```
helrond:
```suggestion
REACT_APP_ENABLE_READING_ROOM_SELECT=true
```
Suggesting changing this to true to reduce confusion, since the string `false` will actually evaluate to a boolean `True` in this implementation.

README.md
```diff
@@ -81,6 +81,10 @@ DIMES uses a [macro implementation](https://lingui.dev/guides/message-extraction
   to present translated strings.
 5. Commit updated code to the GitHub repository.
 
+### Aeon Reading Room Integration
+
+Available dates for reading rooms can be pulled from Aeon via the request broker by setting the `REACT_APP_ENABLE_READING_ROOM_SELECT` environment variable. Not setting this environment variable or leaving it blank will disable this feature.
```
helrond:
```suggestion
Available dates for reading rooms can be pulled from Aeon via the request broker by setting the `REACT_APP_ENABLE_READING_ROOM_SELECT` environment variable. Not setting this environment variable or leaving it blank will disable this feature. Setting this environment variable to any string will activate it.
```
Just being super explicit about this.

.env.deploy
```diff
@@ -3,4 +3,5 @@ REACT_APP_REQUEST_BROKER_BASEURL=${REQUEST_BROKER_BASEURL}
 REACT_APP_LOCALSTORAGE_KEY=DIMESMyList
 REACT_APP_MINIMAP_KEY=DIMESMinimapIntro
 REACT_APP_S3_BASEURL=${AWS_BUCKET_BASEURL}
-REACT_APP_CAPTCHA_SITE_KEY=${CAPTCHA_SITE_KEY}
\ No newline at end of file
+REACT_APP_CAPTCHA_SITE_KEY=${CAPTCHA_SITE_KEY}
+REACT_APP_ENABLE_READING_ROOM_SELECT=false
```
helrond:
```suggestion
REACT_APP_ENABLE_READING_ROOM_SELECT=${ENABLE_READING_ROOM_SELECT}
```
Let's use an environment variable here so we can not implement.

helrond (changes requested):
HI @chryslovelace I am baffled by the behavior in CI. My only thought is that something about the fact that this PR is coming from a different GH organization is causing the Docker build to blow up. 🤷 

I haven't fully tested the reading room functionality because we're not going to be using it in the immediate future so I'll just trust that it works for you. I've added a couple of really minor things here. If you don't mind taking care of those I'll try cloning the development branch in this repository to another branch, repoint this PR at that branch and merge. That *should* allow me to point at development and have CI pass. Sound good?

helrond (approved):

PR #609, opened 2023-06-08
Update BackstopJS request reference file names

### LINKED ISSUE #607: Backstop paths/filenames break Windows build
## Describe the bug
Because of illegal characters in filepaths from Backstop, Windows machines can't checkout branches containing these paths.

## Expected behavior
`yarn serve` should build the site locally with no errors

`error: invalid path 'backstop_data/request_reference/search_facets__api.rockarch.org_facets?category=&limit=40&query=banana__get'`

## Impact on your work
For collaborators using Windows, this blocks local development/testing. 

## Additional context
Two possibilities:
1. See if we can address this in the Backstop configuration (potentially by escaping characters in these paths)
2. Add a new Dockerfile or build stage in the existing Dockerfile which would allow the local container to be used for development.

HaSistrunk:
Fixes #607 

Removes "?" and ":" from filenames to fix issue with Windows forbidden characters in the names. Updates `mock_request.js` to replace the characters.


helrond (approved):
This makes sense to me!

Issue #618, opened 2023-06-22
Review and update loading skeleton styles


HaSistrunk:
## Describe the bug
Loading skeletons do not always have the expected spacing or layout, specifically in the My List modals and on the Agent page. These are not captured in the visual regression testing, so we should comprehensively review all the loading styles.

## To reproduce
1. Add items to your DIMES list
2. Go to the My List page
3. Open the Request Copies modal
4. See that the loading skeletons for possible error messages do not have a sufficient top margin

## Error message or screenshots
My List modal:
<img width="329" alt="image" src="https://github.com/RockefellerArchiveCenter/dimes/assets/15254041/e4191347-4478-4904-b14b-9102b8c41b60">

## Impact on your work
No major impact, but doesn't match the original design of DIMES and doesn't use loading skeletons to their full potential to allow users to anticipate exactly where content may appear when loading is complete.

Issue #615, opened 2023-06-21
Cleanup ModalSearch and Footer line in UI


HaSistrunk:
## Describe the bug
1. The 'Save & Close' button in the Filters modal says 'Save `&amp;` Close'.
2. The h2 of the modal now reads "Filter # Results" instead of "Filter # Search Results" for multiple search results. For one result, it reads "Filter Search Result".
3. The footer has a new `</br>` element after the RAC address and zip code.

## Expected behavior
1. The button should read 'Save & Close'
2. The modal header should read "Filter # Search Results"
3. The address in the footer should not have a break between the street address and zip code.

PR #622, opened 2023-06-23
Fix UI Issues

### LINKED ISSUE #614: Console warning for nesting <Helmet> components on My List page
## Describe the bug
Error in the console in local development states: 
"You may be attempting to nest <Helmet> components within each other, which is not allowed. Refer to our API for more information." This is only occurring on the My List (/list/) page.

## To reproduce
1. Start development server with `yarn start`
2. Open console
3. Go to My List page at /list/
4. See warning

(The warning also comes up in the terminal when you run the BackstopJS tests for the scenarios on the /lists/ page).

## Additional context
This seems to be happening because nesting React components inside `<Helmet>` is not supported (see https://github.com/nfl/react-helmet/issues/326), and in PageMyList the `<Trans>` component is inside of `<Helmet>`:

```      
        <Helmet
          onChangeClientState={(newState) => firePageViewEvent(newState.title)} >
          <Trans comment="Page Title for user's list">
            <title>DIMES: My List</title>
          </Trans>
        </Helmet>
```
Moving `<Helmet>` component inside the `<Trans>` component as it seems we've done on other pages _should_ fix this issue.

HaSistrunk:
Fixes #614 , #615  , #618 

Also fixes incorrect padding in mylist modals select component that was introduced in PR #603.


HaSistrunk:
@helrond I can't get the backstop tests to pass because the error messages in the duplication and reading room request modals aren't loading, so it's just hanging on the loading skeleton. Does it make sense that I'd need to update anything in the backstop configuration after your recent request broker changes?

I tried adding `"readySelector": ".input__error"` to the backstop scenarios for those images so that it would wait to take the image until that class exists, but it just timed out.

helrond:
@HaSistrunk ah, my bad - the data fixtures needed to be updated to account for the new endpoint (`parse-batch`, not `parse`) and the new shape of the data (a list, not a dictionary).

helrond (approved):
Looks good!

PR #625, opened 2023-06-23
Account for cases without search params in Back to Search button on agent pages

### LINKED ISSUE #623: Back to Search from Agent Page Error
## Describe the bug
There is an error that prevents page from loading when trying to use the "Back to Search" button from an agent page when the page was not arrived at through a search. 

## To reproduce
1. Add an item to your list like` /objects/NVTZyHKyg53WnQEKaY3pVY`
2. Go to the My List page
3. Click on the "Found In" collection link for that object. (Nelson A. Rockefeller gubernatorial records, Speeches, Series 33)
4. Click on any of the creator links in the collection details to go to the agent page
5. Click the "Back to Search" button
6. See that the page does not load, and there is an error in the console

## Expected behavior
Good question. I _think_ we'd want that button to take us to a search results page with a search for that agent name.

From console:
image

## Impact on your work
We haven't heard about this error from anyone, but we definitely don't want the site to hang up when someone hits a button.

helrond:
Copies the implementation from RecordsDetail components to programmatically construct the base search URL, accounting for cases where no search parameters are present.

fixes #623

### DIFF
--- a/src/components/PageAgent/index.js
+++ b/src/components/PageAgent/index.js
@@ -227,6 +227,11 @@ const PageAgent = () => {
     }
   }, [wikidata])
 
+   /** Constructs the URL for the 'Back to Search' button */
+   const searchUrl = (
+    params && params.query ? appendParams('/search/', params) : '/'
+  )
+
   if (!found) {
     return (<PageNotFound />)
   }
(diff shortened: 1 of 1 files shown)


HaSistrunk (commented):
Great, working as expected!

PR #634, opened 2023-08-17
Replaces atomic loading skeleton with global loading indicator

### LINKED ISSUE #630: Improve MyList UI to address slow object load times in request modals
Based on reports/requests for support from multiple researchers in Reading Room, it is not always clear to users when objects are loading and need to finish loading in the request modal before they can be selected. Users will try to select checkbox inputs while loading is underway, be unable to select, and think that there is an error.

My testing showed load times to be ~1 min (30 items) to 1 min 45 sec (100 items). 

During multiple tests, I got an error after ~> 1.5 min of loading with between 50-100 objects in my list: `The request to https://request-broker.rockarch.org/api/process-request/parse-batch failed.` Error message: Network error. However, I could not reproduce this consistently so it may have been my personal network connection.

Update the modal UI to show a "Loading..." message (like when the collection content is loading) instead of the skeletons that currently display for any restrictions associated with the object. This solution should be a clearer communication to the user about what is happening.

1+ minutes to load is a long time. Consider redesigning the My List UI overall so that users select which items they will request before opening the request form so that only those objects will load instead of all objects.

helrond:
Removes the not-very-obvious loading skeletons under each item in the list with a global loading indicator. Removes obsolete components and updates tests. Also updates translation files. Fixes #630 

To review this you'll need to jump through a few hoops:
1. Note the IP address of your local machine.
2. Stand up a local instance of Request Broker and append your local IP to the ALLOWED_HOSTS config. Also make sure the AS configs will allow you to connect.
3. Update the`REACT_APP_REQUEST_BROKER_BASEURL` variable in the `.env` file in this repository to be `http://myip/api` where `myip` is your local IP address.
4. Spin up the local DIMES instance by running `yarn start` in this repository.
5. Add a few items to your list.
6. Go to the MyList page and open the Reading Room and Duplication modals and observe the global loading indicator, which disappears once all items in the list have been resolved.


HaSistrunk:
I pushed a suggested tweak to make the text bigger (to the size of our h1 on DIMES), because I think the loading dots appear too small with the standard paragraph text size to notice the dynamic loading indication that they convey. Other than that suggestion, it looks good to me! Let's give it a shot.

image

HaSistrunk (approved):

PR #650, opened 2023-11-28
Adds lingui translations

p-galligan:
Adds different lingui translations.


p-galligan:
Side note, I’m not sure why the `yarn compile` is failing. Maybe because we changed script names in package.json?

README.md
```diff
@@ -30,8 +30,37 @@ To add or update reference images, edit the scenarios in `backstop.json` and run
...
+  * French
+  * German
+  * Italian
+  * Japanese
+  * Korean
+  * Portuguese (Brazilian)
+  * Spanish
+  * Turkish
+
 The repository includes [linguijs](https://lingui.dev/) which is an [Open-source](https://github.com/lingui/js-lingui) Internationalization Framework.  This allows you to do translation work on the UI.
```
helrond: Let's move this up as the first paragraph. 

```suggestion
The repository includes [linguijs](https://lingui.dev/) which is an [Open-source](https://github.com/lingui/js-lingui) Internationalization Framework.  This allows you to translate static UI elements into other languages.
```

README.md
```diff
@@ -30,8 +30,37 @@ To add or update reference images, edit the scenarios in `backstop.json` and run
...
+
+DIMES uses a [macro implementation](https://lingui.dev/guides/message-extraction#macro-usages) for message extraction. This means that all text with a `<Trans></Trans>` block will be automatically extracted for localization files.
+
+You must update the extractions any time new translation objects or languages are found:
+
+1. If adding a new language translation, add a new directory named after the language's ISO 639-1 code to
+  `/src/locales/`. Then add that ISO 639-1 code to the `locales` array in `lingui.config.js`.
+2. Run `$ yarn lingui-extract` to extract translation strings from application code. This should be run every
+  time the HTML structure of the application changes. Failure to run this will result in translated
+  strings being presented as random numbers and letters in the built application. It will automatically generate
```
helrond:
```suggestion
  strings being presented as random numbers and letters in the built application. This command will automatically generate
```

README.md
```diff
@@ -30,8 +30,37 @@ To add or update reference images, edit the scenarios in `backstop.json` and run
...
+  * Spanish
+  * Turkish
+
 The repository includes [linguijs](https://lingui.dev/) which is an [Open-source](https://github.com/lingui/js-lingui) Internationalization Framework.  This allows you to do translation work on the UI.
 
+Lingui requires a configuration file (located in the base directory and named `lingui.config.js`). See the [official documentation](https://lingui.dev/ref/conf) for full instructions on configuration.
+
+DIMES uses a [macro implementation](https://lingui.dev/guides/message-extraction#macro-usages) for message extraction. This means that all text with a `<Trans></Trans>` block will be automatically extracted for localization files.
+
+You must update the extractions any time new translation objects or languages are found:
```
helrond:
```suggestion
You must update files used by lingui any time new translation objects or languages are found:
```

helrond: Wondering if we should break this section up into three different processes:
- Adding a new language
- Updating HTML structure
- Changing translation strings

The commands needed in each case vary, and it might be helpful to have that laid out a bit more clearly. Thoughts?

helrond (changes requested):

README.md
p-galligan: I think that makes sense! I took a stab at this.

helrond:
> Side note, I’m not sure why the `yarn compile` is failing. Maybe because we changed script names in package.json?

Yup, that's probably why. Looks like the Dockerfile needs to be updated. I'm not sure that compile command should actually be in there?

README.md
helrond: Looks good! I think we can get rid of that "You must update files used by lingui any time new translation objects or languages are found:" text now because it's covered by the sections below.

helrond (approved):

PR #660, opened 2024-01-23
Toggle language selection

p-galligan:
Comments out language selection from browser. Defaults to English.


p-galligan:
@helrond This is an attempt to toggle out the automatic language selection and force to English until we're done. Does this work for you?

helrond (approved):
Yup, this should work fine.

PR #661, opened 2024-01-23
Dependency updates and language force

p-galligan:
Updates language selection and dependencies.


p-galligan:
@helrond This is failing because the deploy is failing based on not having AWS S3 access. Is that expected?

helrond:
Yes, that's expected.

p-galligan:
> Yes, that's expected.

Ok, I'll merge to base unless you have any concerns @helrond?

helrond:
Yup, go ahead.

PR #683, opened 2024-03-12
Update DIMES Duplication Request Modal for Fee Changes

### LINKED ISSUE #680: Update DIMES Duplication Request Form to Reflect New Fee Model
With the [updated fee model](https://github.com/RockefellerArchiveCenter/rockarch.org/pull/280), the DIMES request modal needs a refresh.

Update the "fee schedule" link to point to the correct heading in rockarch.org (#duplication-services), since the #duplication-services-and-fee-schedule heading went away.

Instead of just updating the link, we could replace the "note" at the top to say something like:
"We charge a $25 flat fee per item, with a limit of 20 items per year. For details including exceptions for audiovisual materials and special formats, see our [fee schedule](url). For help, email an archivist at archive@rockarch.org."

It's also worth considering if we need the "format" dropdown menu at all, since researchers aren't really choosing a format.

If we decide to change any language, consider the need to update translations.

## Additional Context
Here's what the UI looks like now:
image

### LINKED ISSUE #682: Some times in datepicker are not fully visible
## Describe the bug
Datepicker CSS is causing times to be cut off and not fully in view of user

## To reproduce
1. Go to Reading Room Request modal in my list
2. Click on datepicker and see that times are cut off

image

HaSistrunk:
Fixes #680 
Fixes #682


HaSistrunk:
@p-galligan this will merge some of your recent translation work from development. Is that ok?

p-galligan:
> @p-galligan this will merge some of your recent translation work from development. Is that ok?

Definitely ok! We have the automated language selection commented/toggled off for now, so the language changes won't actually be visible. 

Just please don't manually deploy the dev site. That would mess up review for the last researcher.

PR #703, opened 2024-04-23
Workflows

p-galligan:
Initial test to add GH workflows.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+    branches:
+      - development
+      - base
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    environment:
+      name: development ##Replace with actual environment after testing
```
helrond: Don't forget about this!

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: CI=false yarn run build
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4.0.2
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-to-assume: ${{ secrets.AWS_DEPLOY_ROLE }}
+          role-skip-session-tagging: true
+          role-duration-seconds: 1000
```
helrond: Let's use 900, which is 15 minutes

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - development
+      - base
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    environment:
+      name: development ##Replace with actual environment after testing
```
helrond: Again, don't forget about this!

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Substitute environment variables
+        uses: tvarohohlavy/inplace-envsubst-action@v1.0.0
+        with:
+          files: .env.deploy
+
+      - name: Replace env file
+        run: mv .env.deploy .env
+
+      - name: Build Docker
+        run: docker build .
```
helrond: I think for DIMES this is actually probably part of tests, since we're not actually doing anything with this image, we just want to make sure it builds.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,71 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    environment:
+      name: development ##Replace with actual environment after testing
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
helrond, line 19: I'm sure you've looked at this but can you explain to me why this is necessary?

helrond (changes requested):
Some questions and suggested changes.

.github/workflows/deploy.yml
p-galligan: So, GItHub recommends checking out the code anytime the workflow will interact with the repository's code. "This is an action that checks out your repository onto the runner, allowing you to run scripts or other actions against your code (such as build and test tools). You should use the checkout action any time your workflow will use the repository's code"

Additionally, I believe that unless it's a PR action, GitHub will use the default branch, instead of the working branch.

https://stackoverflow.com/questions/76938677/whats-the-purpose-of-actions-checkoutv3-when-the-repository-is-already-checke

p-galligan: By default, the action will check-out the SHA for that workflow’s event (such as push and pull_request). Otherwise, it uses the default branch.

helrond: Got it, that makes sense.

.github/workflows/test.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          path: ./node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Build Docker
```
helrond: Let's move this to the end. `yarn test` doesn't need this Docker image.

helrond (approved):

PR #706, opened 2024-04-24
Adds additional validation to Request Copies modal

### LINKED ISSUE #704: Limit number of items a user can submit in a duplication request
We are planning to implement request limits for duplication requests in Aeon. To support this change, we should implement some limiting in the DIMES "Request Copies" modal.

When a user fills out the duplication request form and clicks the “Request x Items” button, if there are more than 20 items, a form error appears with the message “Please limit your request to 20 items.” This error will appear in the same way the “No items have been selected to submit” error appears if no items are selected

## Additional context
This will require an update to translations.

helrond:
Adds a check to form validation in the Requests Copies modal to limit requests to 20 items. Also adds translations.

Question: should the request limit be a config value that is set in `.env`?

fixes #704


ctgraham:
> Question: should the request limit be a config value that is set in .env?

Yes, please!  With a conditional allowing it to be unset for unlimited.

helrond:
@ctgraham FYI this only impacts the duplication modal, which I think you all are not currently using? Still, good idea to make it configurable anyway, which I've done here.

HaSistrunk (approved):

Issue #738, opened 2024-09-09
BackstopJS tests not loading object data from mock requests


HaSistrunk:
## Describe the bug
BacktopJS tests that include mock requests to Request Broker are not functioning correctly - object data is not displayed.

## To reproduce
1. Run DIMES locally
2. In another terminal, run the backstopJS tests usuing `yarn backstop-test`
3. See the resulting tests on the My List page and in the request modals that show collection info, but not the associated archival objects.

## Expected behavior
The test snapshots should match the DIMES UI and include the archival object data that has been added to the users list. That data is in the mock requests located in the `backstop_data/request_reference` directory.

## Error message or screenshots
Screenshot of My List test:
<img width="759" alt="image" src="https://github.com/user-attachments/assets/06567b67-ae27-44ae-9cd5-0834894bfa1f">

## Impact on your work
Impacts our ability to the UI for unintended changes.

helrond:
OK I totally rubber ducked myself into figuring this out!

What's happening is that a request is being made behind the scenes to the browser's localStorage, and then that data is used to filter out any items that might have been removed from the list: https://github.com/RockefellerArchiveCenter/dimes/blob/base/src/components/PageMyList/index.js#L264-L266

Because the "browser" storage in this case is empty, the `fetchMyList` function returns an empty list, which then filters out all of the items. 

Solution-wise, I think we need to either find a way to mock that `fetchMyList` function or the browser localStorage state.

Issue #730, opened 2024-08-07
Update Style Library Version


HaSistrunk:
## Is your feature request related to a problem? Please describe.
DIMES is currently importing the RAC style library CSS file version v0.8, but with implementation work in a number of other sites, we've made some updates to the library.

## Describe the solution you'd like
Update the style library version to 0.12.0, and make any needed HTML and local style updates necessary for the update.

## Additional context
There have been some breaking changes that will require some DIMES style tweaks. Check the release notes carefully and run visual regression testing to ensure all visual changes are intended.

PR #739, opened 2024-09-09
Update stylesheet and footer contents

HaSistrunk:
Address issue #730 

Update from RAC style library 0.8.0 to 0.12.0, including associated changes to footer social media icons and updated reading room hours.

Also updates backstop test mock requests to use correct (newer) request broker url. 9 tests are currently failing due to a backstopJS bug that we'll tackle in a separate issue ( #738 ).


helrond (approved):
Looks good!

Issue #732, opened 2024-08-12
Refine search from within collection / object view


ctgraham:
## Is your feature request related to a problem? Please describe.
Archives staff report that they would like patrons to be able to search within a collection view.  Currently, when at the collection / object page rendering, the only opportunity to refine the search is the "Back to Search" button.

## Describe the solution you'd like
Augment the "Back to Search" with a textbox to refine the search on the page itself.

## Describe alternatives you've considered
Archives staff prefer to instruct patrons to "CTRL+F" to find on page, but this is incompatible with dynamic load/scrolling.

## Additional context
This will not preserve the search history, so the meaning of "Back to Search" would change here.

Tagging: @rzhang152

helrond:
Definitely open to hearing solutions here, but wanted to provide a bit of context as to why this functionality currently does not exist in DIMES:
- from a UI perspective, there are some pretty significant challenges in having a search that is global in addition to a search that is scoped to a collection. What do the results pages look like (they usually need to be different)? How does a user know which search they're in? Can you easily get from one context to another?
- Related, as search in DIMES is configured now, search within a collection would not be very helpful because it would only return you one result, which would get you back into the same page you searched from. So the search results would have to be different.
- The addition of the minimap is an attempt to address the use case of needing to find a known item in a collection.

ctgraham:
I acknowledge the complexity of a global search with a scoped search.  I think, though, there can be a balance here.

Consider a search for "Maurine Greenwald", which finds:
https://dimes.rockarch.org/objects/Gxb4UPM2YoB4ybdQRrZNVb?category=&limit=40&query=maurine+greenwald

If I then notice that the Russel Sage Foundation had an interest in the Pittsburgh Survey, a new search highlights this in the current record, but also provides minimap hits to related records:
https://dimes.rockarch.org/objects/Gxb4UPM2YoB4ybdQRrZNVb?category=&limit=40&query=pittsburgh+survey

If "Back to Search" is rebranded "Search Everywhere", you can:
* refine the search locally
* perform a new search globally
* use the back and forward buttons to view search history

![image](https://github.com/user-attachments/assets/8219088c-98c7-42e7-b7ee-d7afc7d04bed)

helrond:
[...]

PR #746, opened 2024-09-25
issue #732: allows users to do a refinement search 

rzhang152:
Give users option to refine a search inside the initial search results on Dimes.


src/components/RecordsDetail/index.js
```diff
@@ -188,6 +193,18 @@ const RecordsDetail = props => {
       label={t({ comment: 'About minimap message', message: 'about minimap' })}
     /> : null
     }
+        <form className ='refine-search-form' >
+        <input type="hidden" name="category" value={curr_category} />
+        <input type="hidden" name="limit" value="40" />
```
ctgraham: Even though this is effectively hard-coded from the previous page, we should still honor whatever `limit` value was passed in to the page, so this `value` should come from the querystring.

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +193,18 @@ const RecordsDetail = props => {
...
+        <label for="query" className='refine-search-label'>Refining Search:</label>
+        <input className='refine-search-input'
+               type='text' 
+               id="query"
+               name="query"
+               placeholder={curr_query || ''}
```
ctgraham: Rather than inserting the current query as a `placeholder`, this should be the field `value`, so that adding an additional term to the query only requires typing the new term, not retyping the original.

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +193,18 @@ const RecordsDetail = props => {
...
+               maxLength={255}
+               size={60}/>
+    </form>
```
ctgraham: I think the form should have a submit button for accessibility purposes.

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +193,18 @@ const RecordsDetail = props => {
       label={t({ comment: 'About minimap message', message: 'about minimap' })}
     /> : null
     }
+        <form className ='refine-search-form' >
+        <input type="hidden" name="category" value={curr_category} />
+        <input type="hidden" name="limit" value="40" />
+        <label for="query" className='refine-search-label'>Refining Search:</label>
```
ctgraham: New user-facing text (such as "Refining Search:") should be wrapped in a [`<Trans>` tag](https://lingui.dev/tutorials/react-patterns) to allow for i18n/l10n.

ctgraham (changes requested):
Hillel's team will be the authority, but some requests from me.

helrond:
I have some higher-level questions about the goals of this PR (aside from the specific things that @ctgraham identified). 

First, it seems like the net effect of this PR would be to remove one click at the expense of masking a lot of search functionality. Users can already get back to the previous search and refine from there by clicking on the "Back to Search" button. I'm not sure that allowing them to update the query only from within a collection view outweighs the additional complexity in code and UI that results. Our user testing indicated that users had no problem identifying and using that "Back to Search" button for search refinement.  

I'm still not sold that a "search within this collection only" is a use case that we are experiencing on our end, so I think if we're going to continue to go down this road this should probably be a configurable feature. That would also allow us to do some testing before we implement to make sure this is something that works for users.

rzhang152:
Thanks for your inputs Hillel.

I think a main purpose of requesting a search refinement feature is to provide the users with a clear picture of their current query, and an options for easily pinning down their search based on the current results they are viewing.

With that in mind, I totally agree that the minimap features allow users to easily click and jump to matches within collections. We also brought your thought and inputs regarding this request during the meeting with our Archives&Special Collections team yesterday. However, from our users' perspectives, we recognize that there are cases where they want to narrow down a search by focusing on a specific collection within the initial results. For instance, a researcher might be looking for a particular name within a specific collection of correspondence found in the search results. In addition, providing option to refine a search within search result page make it more straightforward for users to pin down their finding effectively. The refined search will be conducted on-the-fly and will not directly overwrite the query parameter until execution. This approach aims to minimize the impact on the current minimap and other complex UI features.

I also revised the code based on Clinton's and your input.  Please let me know if any q or feedback. 
Thanks again,

src/components/RecordsDetail/index.js
```diff
@@ -137,6 +137,24 @@ const PanelTextSection = ({ params, text, title }) => {
     (null)
 )}
 
+const RefineSearchSection = ({ refinable, params, title }) => {
+  const curr_category = params && params.category ? params.category : ""
+  return (
+    (refinable && refinable.toLowerCase() === 'true') ?
+    (<form className ='refine-search-form' onSubmit="return false;">
```
helrond: `onSubmit` should return a function, not a string.

src/components/RecordsDetail/index.js
```diff
@@ -137,6 +137,24 @@ const PanelTextSection = ({ params, text, title }) => {
...
+        <input type="hidden" name="category" value={curr_category} />
+        <input type="hidden" name="limit" value={params.limit || '40'}/>
+        <label for="query" className='refine-search-label'>{title}</label>
```
helrond: need to use `htmlFor` instead of `for`

src/components/RecordsDetail/index.js
helrond:
```suggestion
    (<form className='refine-search-form' onSubmit="return false;">
```

src/components/RecordsDetail/index.js
```diff
@@ -137,6 +137,24 @@ const PanelTextSection = ({ params, text, title }) => {
...
+        <input className='refine-search-input'
+               type='search' 
+               id="query"
+               name="query"
```
helrond:
```suggestion
               id='query'
               name='query'
```

src/components/RecordsDetail/index.js
helrond:
```suggestion
        <label for='query' className='refine-search-label'>{title}</label>
```

src/components/RecordsDetail/index.js
```diff
@@ -137,6 +137,24 @@ const PanelTextSection = ({ params, text, title }) => {
     (null)
 )}
 
+const RefineSearchSection = ({ refinable, params, title }) => {
+  const curr_category = params && params.category ? params.category : ""
+  return (
+    (refinable && refinable.toLowerCase() === 'true') ?
+    (<form className ='refine-search-form' onSubmit="return false;">
+        <input type="hidden" name="category" value={curr_category} />
+        <input type="hidden" name="limit" value={params.limit || '40'}/>
```
helrond:
```suggestion
        <input type='hidden' name='limit' value={params.limit || '40'}/>
```

src/components/RecordsDetail/index.js
```diff
@@ -137,6 +137,24 @@ const PanelTextSection = ({ params, text, title }) => {
     (null)
 )}
 
+const RefineSearchSection = ({ refinable, params, title }) => {
+  const curr_category = params && params.category ? params.category : ""
+  return (
+    (refinable && refinable.toLowerCase() === 'true') ?
+    (<form className ='refine-search-form' onSubmit="return false;">
+        <input type="hidden" name="category" value={curr_category} />
```
helrond:
```suggestion
        <input type='hidden' name='category' value={curr_category} />
```

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +204,14 @@ const RecordsDetail = props => {
       label={t({ comment: 'About minimap message', message: 'about minimap' })}
     /> : null
     }
+    <RefineSearchSection
+      refinable={process.env.REACT_APP_REFINE_SEARCH}
+      params={props.params}
+      title={t({
+        comment: 'Placeholder for search refinement textbox',
```
helrond:
```suggestion
        comment: 'Label for search refinement textbox',
```

helrond (changes requested):
A couple other things, mostly very minor.

The translation for the user-facing label will also need to be added as described in the README.

src/components/RecordsDetail/index.js
helrond: I agree with this!

src/components/RecordsDetail/index.js
helrond: +1

HaSistrunk:
Hi there! I'll weigh in with some feedback related to screenreader UX:

So for sighted users, the indication that a refined search has successfully completed is that they might see the minimap squares/links change, and they'll see their new search term in the input. For screenreader users, they'll submit the search and the page will reload with the focus moving to whatever record is in the url. If they've been exploring the collection, that will be somewhere in the collection context tree on the right side of the screen. The screenreader will then start from there since that has focus.

I'm open to ideas, but I'd suggest mitigating that potentially confusing pattern (that doesn't support [WCAG SC 3.2.2](https://www.w3.org/WAI/WCAG22/Understanding/on-input.html)) by adding an aria-live announcement with a message like "Links in minimap navigation have been updated to display new search results". We use an aria-live region already to tell users that collection details on the left side of the page have been updated when someone selects a new record using [react-aria-live](https://www.npmjs.com/package/react-aria-live). You can see that implemented in `components/PageRecords/index.js`. It's not a perfect solution, but folks can at least receive some confirmation feedback after submitting a search and know to navigate back to the minimap to explore the new results.

src/components/RecordsDetail/index.js
helrond: Bumping this - this text is a label, not a placeholder.

src/components/RecordsDetail/index.js
```diff
@@ -137,6 +137,34 @@ const PanelTextSection = ({ params, text, title }) => {
...
+    (refinable && refinable.toLowerCase() === 'true') ?
+    (<form className='refine-search-form'>
+        <input type='hidden' name='category' value={curr_category} />
+        <input type='hidden' name='limit' value={params.limit || '40'}/>
+        <label htmlFor='query' className='refine-search-label'>{title}</label>
+        <input className='refine-search-input'
+               type='search' 
+               id='query'
+               name='query'
+               placeholder={params.query}
```
helrond:
```suggestion
               value={params.query}
```

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +197,33 @@ const RecordsDetail = props => {
       label={t({ comment: 'About minimap message', message: 'about minimap' })}
     /> : null
     }
+    { process.env.REACT_APP_REFINE_SEARCH && 
+      process.env.REACT_APP_REFINE_SEARCH.toLowerCase() === 'true' ?
+      ( <form className='refine-search-form' >
```
helrond:
```suggestion
      ( <form className='refine-search__form' >
```

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +197,33 @@ const RecordsDetail = props => {
...
+        <input type='hidden' name='category' value={props.params.category} />
+        <input type='hidden' name='limit' value={props.params.limit || '40'}/>
+        <Trans comment='Label for search refinement textbox'>
+            <label htmlFor='query' className='refine-search-label'>Refining Search...</label>
```
helrond:
```suggestion
            <label htmlFor='query' className='refine-search__label'>Refining Search...</label>
```

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +197,33 @@ const RecordsDetail = props => {
...
+        </Trans>
+        <input 
+          className='refine-search-input'
+          type='search' 
+          name='query'
+          placeholder={props.params.query} 
+          value={refinedVal.query}
+          onChange={handleChange}/>
+        <Button
+          className='btn btn--orange refine-search-btn'
```
helrond:
```suggestion
          className='btn btn--orange refine-search__btn'
```

src/styles/components/_records-detail.scss
```diff
@@ -73,3 +73,32 @@
...
+  border-radius: 2px;
+  border: 1px solid $silver-grey;
+}
+
+.refine-search-label {
+  display: inline-block;
+  font-weight: bold;
+}
+
+.refine-search-btn {
```
helrond:
```suggestion
.refine-search__btn {
```

src/styles/components/_records-detail.scss
```diff
@@ -73,3 +73,32 @@
...
+  color: inherit;
+  font-size: inherit;
+  height: 35px;
+  width: 146px;
+  background-color: unset;
+  border-radius: 2px;
+  border: 1px solid $silver-grey;
+}
+
+.refine-search-label {
```
helrond:
```suggestion
.refine-search__label {
```

src/styles/components/_records-detail.scss
```diff
@@ -73,3 +73,32 @@
...
 }
+
+// style for search refinement on the search results page
+.refine-search-form {
+  position: relative;
+  width: 200px;
+  padding-top: 20px;
+}
+
+.refine-search-input {
```
helrond:
```suggestion
.refine-search__input {
```

src/styles/components/_records-detail.scss
```diff
@@ -73,3 +73,32 @@
   font-size: inherit;
   font-weight: $font-weight-normal;
 }
+
+// style for search refinement on the search results page
+.refine-search-form {
```
helrond:
```suggestion
.refine-search__form {
```

src/components/RecordsDetail/index.js
```diff
@@ -188,6 +197,33 @@ const RecordsDetail = props => {
...
+    { process.env.REACT_APP_REFINE_SEARCH && 
+      process.env.REACT_APP_REFINE_SEARCH.toLowerCase() === 'true' ?
+      ( <form className='refine-search-form' >
+        <input type='hidden' name='category' value={props.params.category} />
+        <input type='hidden' name='limit' value={props.params.limit || '40'}/>
+        <Trans comment='Label for search refinement textbox'>
+            <label htmlFor='query' className='refine-search-label'>Refining Search...</label>
+        </Trans>
+        <input 
+          className='refine-search-input'
```
helrond:
```suggestion
          className='refine-search__input'
```

helrond (changes requested):
I apologize for missing this earlier, but we've tried to use BEM class names so converting those here. I think I got all of them in my suggestions.

helrond:
Hi @rzhang152 - I was just chatting with @HaSistrunk about the accessibility aspects of this PR. Fundamentally, the challenge here is that the refine search form reloads the page, which has a couple of knock-on implications related to both accessibility and usability:
- Because the page simply reloads instead of going to a search results page, it's hard for users to know what's actually changed. The aria-live implementation, at least as it is currently, won't change that because the announcement is tied to the `onChange` handler of the form, so every time that form value changes screenreader users get an announcement, but they _don't_ get an announcement when the page reloads.
- Related, the focus on page reloads may not be where a user expects it to be - if a user has clicked down into a collection tree and then decides they want to do a refining search, they'll end up back at the same thing they had focused on, which may or may not have a hit associated with it.

After thinking about it, I'd like to suggest the following:
1. I don't think the `aria-live` approach is going to work for this use case without a lot of additional work to store the previous query and then implement a conditional to trigger an announcement when it changes. So, with apologies for leading you down the wrong path, I'd suggest removing that code. Instead I think adding some text in the refine search component which tells users how many hits are in the collection and what the query is, something like "64 hits for 'my query'" would give all users a better indication of what's actually happening. 
2. The page focus could be changed by updating the target of the form to be the top-level collection rather than the object-specific URL, which would return users to the top of the collection tree. However, I think this is an area where user testing and feedback is needed - I'm not sure what users would expect exactly, and what's more or less confusing. This also has accessibility implications for folks using screenreaders.

Let us know how that sounds - we appreciate your thinking and work on this feature!

PR #761, opened 2024-12-03
Only fetches children of collections when they are missing

### LINKED ISSUE #737: Collapse/expand of collection contents accordion creates duplicate child objects
## Describe the bug
Expanding and collapsing collection contents accordion is displaying duplicate child objects of that collection, but only if the user navigates to the collection/object using the minimap.

## To reproduce
1. Search for "ping pong diplomacy"
2. Click on [National Committee on United States-China Relations records result](https://dimes.rockarch.org/collections/6Be4hWNqXCzEwgcsNcR7AU?category=&limit=40&query=ping%20pong%20diplomacy)
3. Click on first minimap square
4. See 2 objects under Series 9
5. Collapse the Series 9 accordian and then expand it
6. See that there are now 4 objects displayed that are duplicates of the original 2 objects in the series.
7. Click on any one of the objects and see that the duplicate objects disappear

*Note that this bug does NOT occur if you do not use the minimap initially to navigate to series 9 (by instead avoiding the minimap and just clicking to expand RG 3 and then Series 9.

## Expected behavior
After navigating to the collection with a specific number of objects, expanding and collapsing the containing collection should have no effect on the contained objects that are displayed.

After clicking on the first minimap square on [this collection contents page](https://dimes.rockarch.org/collections/YHB93QZjHzpyEcxeJbDQqm?category=&limit=40&query=ping%20pong%20diplomacy):
image

After collapsing and re-expanding Ping Pong Diplomacy, Series 9:
image

## Impact on your work
This is a bug we haven't caught before, likely because it is only apparent in collections with very few objects. BUT, it would be very confusing for users when it occurs and could contribute to doubt in system function and accuracy of results.

helrond:
Fixes #737 

@HaSistrunk just looking for a second pair of eyes on this to confirm I haven't introduced another bug in fixing this one.

### DIFF
--- a/src/components/RecordsContent/index.js
+++ b/src/components/RecordsContent/index.js
@@ -179,7 +179,9 @@ export const RecordsChild = props => {
       if (targetIsDirectDescendant) { /* 1 */
         getInitialSiblings(itemUri, params)
       } else { /* 2 */
-        getPages(appendParams(itemUri, {...params, limit: pageSize}))
+        if (!children.length) {
+          getPages(appendParams(itemUri, {...params, limit: pageSize, offset: 0}))
+        }
       }
     }
   }, [isParentCollection, itemUri, params, targetIsDirectDescendant])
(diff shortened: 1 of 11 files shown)


HaSistrunk (approved):
I'm not seeing any issues with this!

PR #768, opened 2024-12-04
Debugs backstop.js implementation

helrond:
- Sets localStorage data
- Adds mocks to additional scenarios
- Adds missing requests

Backstop is still failing for a good number of snapshots, but in general it seems to be because of text spacing and other minor changes.


HaSistrunk (approved):
Amazing, thanks so much for figuring this out! I closely reviewed and then updated the reference images. There are no unintended UI changes that I'm see.

Issue #804, opened 2025-06-24
Improve web analytics tracking approach


HaSistrunk:
## Is your feature request related to a problem? Please describe.
We were using Google Analytics when we first implemented DIMES analytics tracking by adding [a custom script](https://github.com/RockefellerArchiveCenter/dimes/blob/d5581a76a5c963086dbb4137dfcb6a95329176d3/src/components/Helpers.js#L85) to unsure that the analytics get corrrect page title info. However, we are still occasionally getting the generic "DIMES: Online Collections and Catalog of Rockefeller Archive Center" page title in our analytics data even when the page should have a different title.

## Describe the solution you'd like
Update our pageview tracking approach to follow [Matomo's recommendations for tracking single page applications](https://matomo.org/faq/how-to/faq_23636/), which would allow us to remove this helper script in the DIMES code and hopefully better address the incorrect page title issue.

HaSistrunk:
Opening this back up because there's still an issue with capturing pageviews for PageSearch. Matomo hasn't been getting pageview data from the tracking event on this page since 2023, which is when we implemented translation, so it may be related. But we need to diagnose to make sure the logic is solid and reliable across all pages.

HaSistrunk:
[Matomo's recommended option 2 (Use a Custom MTM Event)](https://matomo.org/faq/how-to/capture-the-correct-page-title-in-a-react-spa/) should work for us for all pages with some tweaks:
1. Abstract as hook to use in multiple pages
2. Use `[pathname, title]` instead of Matomo's recommended `[]`, because our pages don't always know the title at mount with API responses. Also pageRecords doesn't remount when you click between records, so records with the same title (somewhat rare but happens) will register.
3. Add `if (!title)` guard so nothing is sent while loading and for when pages hit backend errors

PR #805, opened 2025-06-24
Update custom pageview event for analytics

HaSistrunk:
Address issue #804 

Update: I believe translation is a complicating factor here, so it turns out application code is still required to track pageviews correctly. This PR instead updates the existing code to track URL changes instead of page title changes.

Original PR:
Analytics pageview and page title tracking are now handled in Matomo Tag Manager, with no application code required.

Tested the new configuration in Matomo and it is working as expected, so this PR is just about making sure we remove the obsolete code.

### DIFF
--- a/src/components/Helpers.js
+++ b/src/components/Helpers.js
@@ -81,23 +81,3 @@ export const truncateString = (text, maxLength) => {
     return null
   }
 }
-
-/** Sends a custom pageview event to Matomo Tag Manager.
-* This allows us to ensure that the correct page titles are sent.  */
-var done = false
-var prevTitle = ''
-export const firePageViewEvent = title => {
-  if (title && title !== prevTitle) {
-    done = false
-  }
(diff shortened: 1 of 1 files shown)


helrond (approved):
Always love a PR that removes code!

HaSistrunk:
Going to hold off on merging this for a couple days to confirm that that there is no page title data issue in Matomo, since the issue has always been sporadic.

HaSistrunk:
Sorry to disappoint on removing code, but it turns out there are more complicated things at play in this app that are impacting Matomo's out-of-the-box ability to reliably handle all page titles. 

I also uncovered that our existing code isn't correctly capturing pageview data for pages with translated titles correctly, so I've instead updated our custom pageview event to use URL changes instead of page title changes.

helrond (approved):
Oh well - at last this trigger is better JS...

PR #827, opened 2026-01-13
Use digital object URLs from API

helrond:
Uses digital object URLs available in API data, rather than constructing them using hardcoded parameters.

Two notes:
- At some point we'll likely want to improve this display so that DIMES can display multiple digital objects. This is captured in [another issue](https://github.com/RockefellerArchiveCenter/dimes/issues/756)
- We are still using a configuration to create the download URL for JPEG2000 files. There may be ways to avoid this but the only solutions I could think of involved parsing other URLs, which seems non-ideal because it obfuscates that URL even further.


HaSistrunk:
Ok! This looks good to me. Is there any additional need to account for/make visible errors if the API is down?

helrond:
I don't think so - these changes are just using data that's already been fetched, so any errors in fetching should already be accounted for.

PR #829, opened 2026-01-15
Updates to new version of Mirador

helrond:
This is a pretty substantial update, as it updates the app to React 19 (from 16) and updates a number of other modules in order to support updating of Mirador from v3 to v4.

@HaSistrunk In addition to a general look, there are some sass warnings that show up when running `yarn start` - I'd be interested in any insight you have into those and whether it's anything we need to care about.

Additionally, we'll need to update the visual regression testing snapshots. Among other things I replaced the custom SVG we have for collections (`archive_box`) with the `inventory2` Material Icon.


HaSistrunk:
re: Sass warnings. Yes, I think we will eventually need to care about these, but not immediately if we want to wait. It's just that newer sass and sass-loader versions have breaking changes that do things differently.

1. "The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0" comes from us using an older version of sass-loader (12.20), which is governed by our use of react-scripts 5.0.1 (Create React App) that is itself deprecated. So we may want to modernize things by replacing Create React App, but I'm not sure we need to do that in this PR unless you want to expand it.

2. "Sass @ import rules are deprecated and will be removed in Dart Sass 3.0.0" is a Sass change that we've already [updated in the style library,](https://github.com/RockefellerArchiveCenter/styles/issues/117), but would be a bit of a lift to transition all of the @ imports to @ use, and doesn't need to be done right away either.

HaSistrunk:
Looking at the viewer UI, there are color theme changes that aren't ideal. I'm not 100% what changed in the update to cause that, but maybe we need to look at our configs in PageDigitalObject/index.js to make sure those align with any changes in v4? For example, I'm seeing:
image
 v. the old version 
image


And the collapse sidebar icon isn't visible anymore: 
image
 v. the older version:
image

helrond:
Thanks @HaSistrunk - I was able to resolve the button color issue by tweaking the config, as you suggested.

HaSistrunk (approved):
Great! Looks good to me. I just tweaked the tile icon spacing styles to make them consistent between icons and the backstop scenarios that reference downshift selectors (that apparently changed a little in the update).

Issue #841, opened 2026-02-26
Add format notes for document type records to UI


HaSistrunk:
## Is your feature request related to a problem? Please describe.
We currently display `physdesc` and `materialspec` notes in the "Format" section of record details in the UI for moving image, photograph, and audio, filtering out document type records (see [RecordsDetail lines 54-75](https://github.com/RockefellerArchiveCenter/dimes/blob/7a610ed11099c572869a949976d4afeba7762c21/src/components/RecordsDetail/index.js#L54)). However, archives staff have indicated interest in also including these notes for rolled maps (or presumably other non-normative document formats).

## Describe the solution you'd like
Remove the filter so that the format displays these notes for all records, regardless of type.

## Describe alternatives you've considered
Add filtering so that just "documents" would still not appear, but all `physdesc` and `materialspec` notes appear even if they are documents.

## Additional context
This addition would likely help researchers better understand the record context and possible contents, and also if the is in a  format that would require different procedures to access or digitize.

We originally decided to filter out document types most likely because we only wanted to display the "exceptions" to the default. Without the filter, most records would display: "Format: documents", which adds UI clutter.

### Example of current UIs
Documents:
<img width="456" height="114" alt="Image" src="https://github.com/user-attachments/assets/871ec49b-2ae1-4617-afec-ee9e45855908" />

Photographs:
<img width="452" height="113" alt="Image" src="https://github.com/user-attachments/assets/f3d27952-21d7-4569-b984-8fad72b60466" />

Moving image:
<img width="451" height="130" alt="Image" src="https://github.com/user-attachments/assets/2663f16e-9dcf-4603-8f6f-091f9bdbaa61" />

### Example of UIs without format filters

Documents (standard):
<img width="446" height="109" alt="Image" src="https://github.com/user-attachments/assets/e16bb71f-0507-459d-95fd-198694a277dd" />

Documents with notes (rolled map example):

<img width="454" height="212" alt="Image" src="https://github.com/user-attachments/assets/5a0540ac-0b7b-4c6b-82c4-6140126172c0" />

HaSistrunk:
The one issue I see in removing the filter entirely is that at the collection level is that I'm seeing some misleading and not useful description. Examples:

<img width="461" height="344" alt="Image" src="https://github.com/user-attachments/assets/247b6fed-78b0-4ea3-aa8c-81975269f493" />


Only "documents" when there are other formats in the collection.
[...]

PR #842, opened 2026-02-26
Remove document type records filter for format in record details

HaSistrunk:
Fix issue #841

Remove existing filter that only displayed format information for photographs, audio, and moving image records to be able to also display "Material Specific Details" and "Physical Description" notes for maps, flat files, and other formats associated with documents.

One question to resolve:
Is it a problem at the collection level to display all these values unfiltered? See [examples of unnecessary or misleading formats info](https://github.com/RockefellerArchiveCenter/dimes/issues/841#issuecomment-3968482799), such as the Nelson Rockefeller papers displaying "Formats: documents" when there are many other formats contained in the collection. Perhaps this is a description issue to resolve separately?


helrond (approved):
This looks right to me. 

In terms of the misleading formats at the collection level, we currently derive formats other than "documents" from subjects applied directly to resource record (or archival object) so yeah, that feels like a description thing to me.

